PR #50, opened 2026-01-27
Improve instructions around integration tests

astrojuanlu:
## Issue

After running our integration tests dozens of times, I think I'm ready to backport https://github.com/canonical/mysql-k8s-operator/pull/679

I submitted this against 8.4/edge, in the understanding that this will soon become the default branch, so the most visible in the repo. Trying to minimize changes in 8.0/edge from this point.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ but you will need a full local setup depending on the substrate.
 Alternatively, you can use [spread], which will take care of that for you.
 For example, to run a single test with spread, do
 
-```
+```shell
 ~/go/bin/spread lxd-vm:ubuntu-24.04:tests/spread/integration/test_database.py:juju36
 ```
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -40,14 +40,25 @@ You can create an environment for development:
...
 ```
 
+To run integration tests, you can run `tox run -e integration`,
+but you will need a full local setup depending on the substrate.
+Alternatively, you can use [spread], which will take care of that for you.
+For example, to run a single test with spread, do
+
+```
+~/go/bin/spread lxd-vm:ubuntu-24.04:tests/spread/integration/test_database.py:juju36
+```
```
sinclert-canonical: Should we:

1. Set the type of this code-block (i.e. `shell`)?
2. Prepend the command with the recommended way to install spread?

sinclert-canonical (approved):

CONTRIBUTING.md
astrojuanlu: I added the `shell`. I'd rather not tie ourselves to the spread installation instructions.

PR #62, opened 2026-02-02
[MISC] 8.4 - Remove pytest-operator plugin

sinclert-canonical:
This PR removes the installation of `pytest-operator` (and `pytest-asyncio` when present) from the operators.

This was a leftover from the migration to Jubilant, where such plugin is not valid anymore (see [migration steps](https://github.com/canonical/charmIng-recipes/blob/main/integration-testing/commands/migrate-jubilant.md)). Every instance of the _abort-on-fail_ plugin strategy has been replaced by the pytest `-x / --exitfirst` flag.


sinclert-canonical:
Skip-pre-upgrade tests fail because of an un-related reason: something changed on 23/01 tweaking with unit statuses.

Merging it as is.

PR #67, opened 2026-02-05
[DPE-9035] Validate passwords

astrojuanlu:
## Issue

First attempt at validating the passwords. Working for Kubernetes and Machines. Requires a dev version of mysql-shell-client.

Had to get creative to guarantee 100 % probability of `generate_random_password` yielding a compliant one, but other than that I'm happy with the approach.

## Checklist


astrojuanlu:
I've tried a few other things but they don't work. Realistically, the only way to execute `INSTALL COMPONENT` is to wait until initialization is mostly complete.

I'll leave this in draft a bit longer but I'll proceed translating this procedure to the machine charm.

kubernetes/src/charm.py
```diff
@@ -761,6 +771,11 @@ def _configure_instance(self, container) -> None:
                 self._mysql.install_plugins(["audit_log"])
             self._mysql.install_plugins(["binlog_utils_udf"])
 
+            # TODO: Is there a better way rather than having to restart again?
```
astrojuanlu: Possibly, but I'm less confident now.

kubernetes/src/utils.py
```diff
@@ -16,6 +16,9 @@ def generate_random_password(length: int) -> str:
     Returns:
         A randomly generated string intended to be used as a password.
     """
+    # FIXME: There is a nonzero probability that the password doesn't have
+    # any numbers, or any uppercase letters, or any lowercase letters,
+    # in which case validation will fail
```
astrojuanlu: This is a real concern but I didn't want to touch the generator function just yet before having some security advice. I thought of different possible solutions, none of them clean.

machines/src/mysql_vm_helpers.py
```diff
@@ -403,6 +408,30 @@ def reset_root_password_and_start_mysqld(self) -> None:
...
+            raise MySQLWritePasswordValidationConfigAndStartMySQLDError(
+                "Failed to restart mysqld"
+            ) from e
+
+        try:
+            self.wait_until_mysql_connection()
+        except MySQLServiceNotRunningError as e:
+            raise MySQLWritePasswordValidationConfigAndStartMySQLDError(
+                "mysqld service not running"
+            ) from e
```
astrojuanlu: Took me a long while to figure out this part, I was getting connection errors and the charm getting stuck in Blocked status. There's bad observability on some parts of the code and it's a combination of subprocess outputs being suppressed, the way logging and error handling is done... no specific improvements yet, but I can definitely say that the debugging experience here was bad.

machines/src/utils.py
```diff
@@ -16,6 +16,9 @@ def generate_random_password(length: int) -> str:
     Returns:
         a string with random letters and digits of length specified
     """
+    # FIXME: There is a nonzero probability that the password doesn't have
+    # any numbers, or any uppercase letters, or any lowercase letters,
+    # in which case validation will fail
```
astrojuanlu: Same as above

kubernetes/src/utils.py
astrojuanlu: Ended up adding a `while` loop that waits until the generated password passes the checks. Not exactly elegant but it works.

sinclert-canonical:
> Things I've tried that are not reflected in this PR:
> - set_instance_variable ❌ changes are not persistent across restarts.

Have you tried calling that method with the `scope.PERSIST` argument ([reference](https://github.com/canonical/mysql-shell-client/blob/1387d767fb56b5a6d6d9924cab092aa90aa75781/src/mysql_shell/models/statement.py#L21-L30))?

astrojuanlu:
> Like [SET GLOBAL](https://dev.mysql.com/doc/refman/8.4/en/set-variable.html), [SET PERSIST](https://dev.mysql.com/doc/refman/8.4/en/set-variable.html) sets the global variable runtime value, but also writes the variable setting to the mysqld-auto.cnf file (replacing any existing variable setting if there is one).

I had not. Thanks for the pointer @sinclert-canonical ! This will be a bit simpler, and I think it will save us the restart.

astrojuanlu:
Done @sinclert-canonical , please have another pass

sinclert-canonical:
> Done @sinclert-canonical , please have another pass

Will review before EOD. Please update PR description accordingly.

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -1376,6 +1392,32 @@ def uninstall_plugins(self, plugins: list[str]) -> None:
...
+                    logger.warning(f"{component=} file not found. Skip installation")
+                    continue
+                if component in installed_components:
+                    logger.info(f"{component=} already installed")
+                    continue
+
+                try:
+                    self._instance_client_tcp.install_instance_component(component)
+                except ExecutionError as e:
+                    raise MySQLPluginInstallError() from e
```
sinclert-canonical: This whole block could be de-indented if the first conditional is inverted:

```python
if component in installed_components:
    continue
...
```

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -1376,6 +1392,32 @@ def uninstall_plugins(self, plugins: list[str]) -> None:
...
+        installed_components = self._instance_client_tcp.search_instance_components("%")
+
+        for component in components:
+            if component not in installed_components:
+                if component not in ALLOWED_COMPONENTS:
+                    logger.warning(f"{component=} is not supported")
+                    continue
+                if not self._plugin_file_exists(ALLOWED_COMPONENTS[component]):
+                    logger.warning(f"{component=} file not found. Skip installation")
+                    continue
```
sinclert-canonical: We do not need to check whether the component exists so far, given that the allowed component you are listing is available from the very first version of the MySQL 8.4 charmed snap we have released. I would rather add it when it is really needed.

This check is present in the `install_plugin` counterpart method (see [code](https://github.com/canonical/mysql-operators/blob/43fb89e5cf93e38a0a9389300e84e48dc4327dfe/kubernetes/lib/charms/mysql/v0/mysql.py#L1339-L1341)) as some of the initial revisions of the MySQL 8.0 charmed snap did not have some of the allowed plugins.

kubernetes/src/charm.py
```diff
@@ -761,6 +761,9 @@ def _configure_instance(self, container) -> None:
                 self._mysql.install_plugins(["audit_log"])
             self._mysql.install_plugins(["binlog_utils_udf"])
 
+            self._mysql.install_components(["file://component_validate_password"])
+            self._mysql.persist_password_validation_configuration()
```
sinclert-canonical: If the `validate_password` component is always going to be installed, we may as well put the configuration values directly in the mysqld config (i.e. in [this section](https://github.com/canonical/mysql-operators/blob/43fb89e5cf93e38a0a9389300e84e48dc4327dfe/kubernetes/lib/charms/mysql/v0/mysql.py#L1156-L1177)), prefixing them with `loose-`, so that MySQL does not crash due to unknown configuration options.

That way we will:
- Avoid creating an additional charmlib method.
- Prevent opening N new connections to MySQL Server, upon starting the charm, one per config option.

kubernetes/src/utils.py
```diff
@@ -17,7 +17,16 @@ def generate_random_password(length: int) -> str:
...
-    return "".join([secrets.choice(choices) for _ in range(length)])
+    # Might seem risky but in fact the probability that a password doesn't pass these checks is low
+    while True:
+        password = "".join([secrets.choice(choices) for i in range(length)])
+        # These checks are consistent with our rules for the password validation MySQL component
+        if (
+            any(c.islower() for c in password)
+            and any(c.isupper() for c in password)
+            and any(c.isdigit() for c in password)
+        ):
```
sinclert-canonical: Non blocking suggestion:

```python
if all((
    any(c.islower() for c in password),
    any(c.isupper() for c in password),
    any(c.isdigit() for c in password),
)):
```

sinclert-canonical (commented):
A couple of general comments:

- If I am not mistaken, the mysql-provider interface needs to be updated as well (see [code](https://github.com/canonical/mysql-operators/blob/43fb89e5cf93e38a0a9389300e84e48dc4327dfe/kubernetes/src/relations/mysql_provider.py#L71-L72)), so that if the user provides a custom password it is validated as well. Is this validation performed by MySQL server regardless?
- I have left comments only in the Kubernetes implementation, but the same ones are applicable to the VM one.

paulomach (commented):
@astrojuanlu why don't do validation at the charm level only, without the component?
When using user secrets, we are having a set of credentials. If we choose to update the set atomically, it will be simpler to validate it's content at the charm before db submission, so to avoid partial application of credentials.
I am making sense?

kubernetes/src/charm.py
astrojuanlu: TIL https://www.percona.com/blog/using-the-loose_-option-prefix-in-my-cnf/

Definitely simplifies everything. Will switch to doing this.

kubernetes/lib/charms/mysql/v0/mysql.py
astrojuanlu: So you're saying I remove the `self._plugin_file_exists` check? If so, fair enough, although I'll leave the code commented out so that the decision is easier to remember

astrojuanlu:
> When using user secrets, we are having a set of credentials. If we choose to update the set atomically, it will be simpler to validate it's content at the charm before db submission, so to avoid partial application of credentials.  I am making sense?

As far as I understand, the moment we get to Juju secrets, it's more or less the same to validate the password on the charm side than trying to set it and catching the exception, right?

> why don't do validation at the charm level only, without the component?

Seemed more robust to me to use the native MySQL functionality rather than duplicating it on the Python side. Also can potentially make it easier for the user to tweak the rules to their liking.

paulomach:
> > When using user secrets, we are having a set of credentials. If we choose to update the set atomically, it will be simpler to validate it's content at the charm before db submission, so to avoid partial application of credentials.  I am making sense?
> 
> As far as I understand, the moment we get to Juju secrets, it's more or less the same to validate the password on the charm side than trying to set it and catching the exception, right?

Make sense. I was unsure about multiple user DDL atomicity, but they can be handled in a single transaction [ref](https://dev.mysql.com/blog-archive/mysql-8-0-making-user-management-ddls-atomic/)


> 
> > why don't do validation at the charm level only, without the component?
> 
> Seemed more robust to me to use the native MySQL functionality rather than duplicating it on the Python side. Also can potentially make it easier for the user to tweak the rules to their liking.

👍🏻

astrojuanlu:
> If I am not mistaken, the mysql-provider interface needs to be updated as well (see [code](https://github.com/canonical/mysql-operators/blob/43fb89e5cf93e38a0a9389300e84e48dc4327dfe/kubernetes/src/relations/mysql_provider.py#L71-L72)), so that if the user provides a custom password it is validated as well. Is this validation performed by MySQL server regardless?

If I understand that code correctly, either the password is stored in the relation data (coming from an already-validated set-password attempt), or a new one is generated. So no further validation is needed. Am I missing something?

kubernetes/pyproject.toml
```diff
@@ -24,7 +25,7 @@ charm-libs = [
     # tempo_coordinator_k8s/v0/charm_tracing.py requires pydantic
     "pydantic~=1.10",
     # mysql/v0/*.py"
-    "mysql_shell_client~=0.6",
+    "mysql-shell-client @ git+https://github.com/canonical/mysql-shell-client.git@juanlu/install-components",
```
paulomach: Let's release first

astrojuanlu:
> > If I am not mistaken, the mysql-provider interface needs to be updated as well (see [code](https://github.com/canonical/mysql-operators/blob/43fb89e5cf93e38a0a9389300e84e48dc4327dfe/kubernetes/src/relations/mysql_provider.py#L71-L72)), so that if the user provides a custom password it is validated as well. Is this validation performed by MySQL server regardless?
> 
> If I understand that code correctly, either the password is stored in the relation data (coming from an already-validated set-password attempt), or a new one is generated. So no further validation is needed. Am I missing something?

Fixing locally.

But also, backup tests are failing. Looking into it.

kubernetes/charmcraft.yaml
```diff
@@ -52,7 +52,7 @@ parts:
     source: .
     after:
       - poetry-deps
-    poetry-export-extra-args: ['--only', 'main,charm-libs']
```
paulomach, line 55: revert after package release

machines/charmcraft.yaml
```diff
@@ -67,6 +67,8 @@ parts:
         apt-get install rustup -y
       fi
 
+      apt-get install git -y
```
paulomach: Revert when using released package

paulomach (dismissed):
Looks good. It will be simple to add user secret support. We need to release the mysql-shell-client package before merging though

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -1376,6 +1399,31 @@ def uninstall_plugins(self, plugins: list[str]) -> None:
...
+            if component not in ALLOWED_COMPONENTS:
+                logger.warning(f"{component=} is not supported")
+                continue
+
+            # Since we're checking ALLOWED_COMPONENTS already,
+            # we do not check for file existence
+
+            if component in installed_components:
+                logger.info(f"{component=} already installed")
+                continue
```
sinclert-canonical: This check is duplicated with the first one within the loop, a few lines above.

kubernetes/src/relations/mysql_provider.py
```diff
@@ -168,6 +146,28 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
...
+        self.charm._mysql.update_endpoints(DB_RELATION_NAME)
+
+        # create k8s services for endpoints
+        self.charm.k8s_helpers.create_endpoint_services(["primary", "replicas"])
+
+        primary_endpoint = dotappend(socket.getfqdn(f"{self.charm.app.name}-primary"))
+        replicas_endpoint = dotappend(socket.getfqdn(f"{self.charm.app.name}-replicas"))
+
+        # wait for endpoints to be ready
+        self.charm.k8s_helpers.wait_service_ready((primary_endpoint, 3306))
```
sinclert-canonical: I would probably leave these lines **before** the `try-except` block, as they seem to be related to networking, and not with _setting Juju databag values if everything goes well_. In fact, the first line is  placed before the block in the VM counterpart.

Just move them outside of the `try-except` block.

machines/lib/charms/mysql/v0/mysql.py
```diff
@@ -1376,6 +1399,31 @@ def uninstall_plugins(self, plugins: list[str]) -> None:
...
+            if component not in ALLOWED_COMPONENTS:
+                logger.warning(f"{component=} is not supported")
+                continue
+
+            # Since we're checking ALLOWED_COMPONENTS already,
+            # we do not check for file existence
+
+            if component in installed_components:
+                logger.info(f"{component=} already installed")
+                continue
```
sinclert-canonical: This check is duplicated with the first one within the loop, a few lines above.

sinclert-canonical (commented):
LGTM overall. Waiting for a package release + couple of adjustement.

astrojuanlu:
Dropped the commits that installed the dev version of mysql-shell-client, I'll restart the tests after 0.7 is released

kubernetes/tests/integration/integration/test_password_validation.py
```diff
@@ -0,0 +1,135 @@
...
+    try:
+        juju.run(
+            unit=primary_unit_name,
+            action="set-password",
+            params={"username": CLUSTER_ADMIN_USERNAME, "password": short_password},
+        )
+    except TaskError as e:
+        logger.info(f"Action failed as expected with exception: {e}")
+    else:
+        raise AssertionError("set-password task should have failed")
```
sinclert-canonical: We typically have been testing fail-conditions with the provided Pytest decorator:

```python
with pytest.raises(TaskError):
    juju.run(
        ...
    )
```

(Same comment for the same snippet +/- 30 LOC below).

machines/tests/integration/integration/test_password_validation.py
```diff
@@ -0,0 +1,133 @@
...
+    try:
+        juju.run(
+            unit=primary_unit_name,
+            action="set-password",
+            params={"username": CLUSTER_ADMIN_USERNAME, "password": short_password},
+        )
+    except TaskError as e:
+        logger.info(f"Action failed as expected with exception: {e}")
+    else:
+        raise AssertionError("set-password task should have failed")
```
sinclert-canonical: We typically have been testing fail-conditions with the provided Pytest decorator:

```python
with pytest.raises(TaskError):
    juju.run(
        ...
    )
```

(Same comment for the same snippet +/- 30 LOC below).

sinclert-canonical (dismissed):

sinclert-canonical (approved):
Great work! :shipit:

paulomach (approved):
Great stuff

PR #77, opened 2026-02-10
[Misc] Remove offending upgrade-skip test

paulomach:
## Issue

skip pre-upgrade-check is flaky and test an unsupported use case

## Solution

Remove it


sinclert-canonical (approved):
🔥

astrojuanlu (approved):

PR #88, opened 2026-02-13
feat: Use s3-integrator from track 2/ with the newer object-storage lib

theoctober19th:
## Proposal
We'd like to update the S3 lib to v1 (now called object_storage lib).

In our integration tests, we'd like to use `s3-integrator` from track `2/`, with new features like support for secret configuration, and also sharing data over the relation using secrets.

## Checklist


.github/workflows/ci.yaml
```diff
@@ -116,3 +116,25 @@ jobs:
       artifact-prefix: ${{ needs.build.outputs.artifact-prefix }}
       path-to-charm-directory: ${{ matrix.path }}
     secrets: inherit
+
+  release-test-branch:
```
sinclert-canonical: Same as other PR reviews: this CI step needs to go before merging.

kubernetes/tests/integration/integration/helpers_backups.py
```diff
@@ -53,6 +55,14 @@ def local_tmp_folder(name: str = "tmp"):
     shutil.rmtree(tmp_folder)
 
 
+def prepare_s3_credentials_secret(juju: Juju, credentials: dict[str, str]) -> str:
+    """Prepare the s3 credentials secret and return the secret URI."""
+    secret_name = str(uuid.uuid4())[:8]
+    secret_uri = juju.add_secret(secret_name, content=credentials)
+    juju.grant_secret(secret_uri, S3_INTEGRATOR)
+    return secret_uri
```
sinclert-canonical: **This is the 1st occurrence of this snippet. Feedback applies to all occurrences.**

I see this helper function is doing a very _general purpose_ operation, despite some of its names. I think it can be defined in the `tests/integration/helpers_ha.py` module as follows:

```python
def create_app_secret(juju: Juju, app_name: str, contents: dict[str, str]) -> str:
    """Creates an application secret and return the secret URI."""
    secret_name = str(uuid.uuid4())[:8]
    secret_uri = juju.add_secret(secret_name, content=contents)
    juju.grant_secret(secret_uri, app_name)
    return secret_uri
```

kubernetes/tests/integration/integration/helpers_backups.py
```diff
@@ -116,12 +126,8 @@ def build_and_deploy_operations(
...
     )
     juju.config(S3_INTEGRATOR, cloud_configs)
-    s3_unit_name = get_app_units(juju, S3_INTEGRATOR)[0]
-    juju.run(
-        unit=s3_unit_name,
-        action="sync-s3-credentials",
-        params=cloud_credentials,
-    )
+    secret_uri = prepare_s3_credentials_secret(juju, cloud_credentials)
+    juju.config(S3_INTEGRATOR, {"credentials": secret_uri})
```
sinclert-canonical: **This is the 1st occurrence of this snippet. Feedback applies to all occurrences.**

Nit: let's do configuration altogether.

```diff
- juju.config(S3_INTEGRATOR, cloud_configs)
- secret_uri = prepare_s3_credentials_secret(juju, cloud_credentials)
- juju.config(S3_INTEGRATOR, {"credentials": secret_uri})

+ secret_uri = prepare_s3_credentials_secret(juju, cloud_credentials)
+ juju.config(S3_INTEGRATOR, cloud_configs)
+ juju.config(S3_INTEGRATOR, {"credentials": secret_uri})
```

kubernetes/pyproject.toml
```diff
@@ -10,6 +10,7 @@ main = [
     "ops[tracing]~=3.5",
     "pyyaml~=6.0",
     "tenacity~=8.2",
+    "object-storage-charmlib~=0.1.0",
```
sinclert-canonical: Nit: let's keep dependency ordering alphabetically.

sinclert-canonical (commented):
🧡 Thanks for helping us keep our charm up-to-date!

Left a few comments.

astrojuanlu (approved):
Apart from what @sinclert-canonical said, LGTM! Great to see one less charm lib 😉 And the changes are minimal enough that this was easy to review. Well done!

kubernetes/tests/integration/integration/helpers_backups.py
theoctober19th: ~~Thanks for the comment. I see there's another module `helpers_backups.py` -- perhaps it's better to keep this utility function there instead of `helpers_ha`, since this is going to be used in the backup tests?~~

NVM -- I saw some similar utility functions in `helpers_ha`, I'll keep this new one there as well.

theoctober19th:
Hi @sinclert-canonical, I've addressed your comments, can I have another round of review?

theoctober19th:
I believe the next task we should do here is now:
1. to update the MySQL 8.4/edge documentation so that the docs mention deploying `s3-integrator` from proper channel (2/edge) and also the correct way to configure it (using Juju secret instead of the juju action previously). 
2. to update the Terraform product / solution module to use the s3-integrator from proper channel, and also expose TF variables to plug into the charm configuration. And perhaps update the docs for the terraform deployment.

Should we perhaps plan for that now? 

cc: @astrojuanlu @sinclert-canonical @paulomach @a-velasco

sinclert-canonical:
> To update the Terraform product / solution module to use the s3-integrator from proper channel

Would these changes require to use a S3 Integrator from the `2/*` channels? or `1/*` ones would still be compatible? If the former, documentation + Terraform changes should be proposed and left _ready to merge_ in order to reduce the de-sync time between code and docs.

> Should we perhaps plan for that now?

Absolutely. cc @paulomach

paulomach:
> > To update the Terraform product / solution module to use the s3-integrator from proper channel
> 
> Would these changes require to use a S3 Integrator from the `2/*` channels? or `1/*` ones would still be compatible? If the former, documentation + Terraform changes should be proposed and left _ready to merge_ in order to reduce the de-sync time between code and docs.
> 
> > Should we perhaps plan for that now?
> 
> Absolutely. cc @paulomach

Created [two tasks](https://warthogs.atlassian.net/issues/?jql=issueKey%20in%20(DPE-9722%2CDPE-9724))

As this are affecting 8.4, it's not a big issue to have some desync before releasing to stable. I will try to schedule the tasks asap though, as long it don't affect roadmap

paulomach (approved):

sinclert-canonical:
The MySQL _product_ Terraform module is defined somewhere else (see [K8s definitions](https://github.com/canonical/mysql-k8s-bundle/tree/main/terraform) / [VM definitions](https://github.com/canonical/mysql-bundle/tree/main/terraform)), but the docs on how to integrate with S3 Integrator are defined **in this same repository**. I find it a little wasteful to update the code and the docs in separated PRs unless the original PR blew out of proportion, which I do not think it did.

theoctober19th:
> Would these changes require to use a S3 Integrator from the 2/* channels? or 1/* ones would still be compatible? If the former, documentation + Terraform changes should be proposed and left ready to merge in order to reduce the de-sync time between code and docs.

While the compatibility is supported (and tested in s3-integrator), the de-sync is not recommended, if possible. I believe the updates in the docs don't require much effort since the only change is the channel and the way the configuration is done. (Perhaps the TF modules update would require a little bit more).

sinclert-canonical:
I will go ahead a include the documentation changes myself

sinclert-canonical (approved):

PR #89, opened 2026-02-13
[MISC] integration test fixes

paulomach:
This PR proposes couple fixes from observed behavior:
* add missing `trust` on k8s - though not always relevant, keeps it consistent
* add architecture constraints to router/test-app - when missing, juju might select wrong revision/arch on k8s - resulting in test timeout due to app pod(s) never being scheduled, as it will have a architecture bound node affinity rule
* increased default `juju.wait` calls delay from 1sec to 2sec - mitigate test failure due cli errors, more present since juju 3.6.14
* adds a sleep between some `juju.deploy` and `juju.wait` - mitigate the infamous `filesystem for storage instance "database/x" not found` error
* 
* ~~attempt to remove the `retry_on_cli_error` wrapper on k8s multi relations test~~
* ~~and last but no least, fix test_spaced_db by setting bases for the isolated mysql-test-app, as incorrect bases is pick when not set.~~

Generally, I've observed that failed tests on first run decreased from ~25 to ~15, with the remaining failures related to the cli errors, some setup errors and the remaining, due to timeouts.

### DIFF
--- a/machines/tests/integration/integration/spaces/test_spaced_db.py
+++ b/machines/tests/integration/integration/spaces/test_spaced_db.py
@@ -98,6 +98,8 @@ def test_integrate_with_isolated_space(juju: Juju):
         isolated_app_name,
         constraints={"spaces": "isolated"},
         bind={"database": "isolated"},
+        num_units=1,
+        base="ubuntu@22.04",
         channel="latest/edge",
     )
     juju.wait(
--- a/machines/tests/integration/integration/spaces/test_spaced_db.py
+++ b/machines/tests/integration/integration/spaces/test_spaced_db.py
@@ -48,10 +48,12 @@ def test_build_and_deploy(juju: Juju, lxd_spaces, charm) -> None:
     juju.wait(
         ready=wait_for_apps_status(jubilant_backports.all_active, DATABASE_APP_NAME),
         timeout=TIMEOUT,
+        delay=2,
     )
     juju.wait(
         ready=wait_for_apps_status(jubilant_backports.all_waiting, APPLICATION_APP_NAME),
         timeout=TIMEOUT,
+        delay=2,
     )
 
 
@@ -64,6 +66,7 @@ def test_integrate_with_spaces(juju: Juju):
     juju.wait(
         ready=jubilant_backports.all_active,
         timeout=TIMEOUT,
+        delay=2,
     )
 
     unit = get_app_units(juju, APPLICATION_APP_NAME)[0]
@@ -82,6 +85,7 @@ def test_integrate_with_spaces(juju: Juju):
     juju.wait(
         ready=lambda status: APPLICATION_APP_NAME not in status.apps,
         timeout=TIMEOUT,
+        delay=2,
     )
 
 
@@ -99,6 +103,7 @@ def test_integrate_with_isolated_space(juju: Juju):
     juju.wait(
         ready=wait_for_apps_status(jubilant_backports.all_waiting, isolated_app_name),
         timeout=TIMEOUT,
+        delay=2,
     )
 
     # Relate the database to the application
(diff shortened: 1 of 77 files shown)


machines/tests/integration/integration/spaces/test_spaced_db.py
```diff
@@ -94,11 +98,14 @@ def test_integrate_with_isolated_space(juju: Juju):
         isolated_app_name,
         constraints={"spaces": "isolated"},
         bind={"database": "isolated"},
+        num_units=1,
+        base="ubuntu@22.04",
```
sinclert-canonical, line 102: For history purposes:

As discussed with Paulo, this is the most critical fix of the whole PR. With the recent release of `mysql-test-app` 24.04 base, Jubilant is not smart enough to select the correct revision, so this application fails to get into a ready state.

sinclert-canonical (commented):
Let's try to get rid of this K8s integration tests [snowflake retry-block](https://github.com/canonical/mysql-operators/blob/f18bfc035e031c93b557a19b45ba81f74a61469f/kubernetes/tests/integration/integration/test_multi_relations.py#L139-L154) by leveraging the proposed `delay` argument when doing a Jubilant wait, and merge these changes ASAP.

astrojuanlu:
> Let's try to get rid of this K8s integration tests [snowflake retry-block](https://github.com/canonical/mysql-operators/blob/f18bfc035e031c93b557a19b45ba81f74a61469f/kubernetes/tests/integration/integration/test_multi_relations.py#L139-L154) by leveraging the proposed delay argument when doing a Jubilant wait, and merge these changes ASAP.

Notice that the changes are not equivalent. `juju.wait` does have a `delay` parameter, but every CLI invokation is susceptible of failing.

In fact, with these changes, we see `test_multi_relations.py` is failing:

- https://github.com/canonical/mysql-operators/actions/runs/22079079841/job/63813131384?pr=89 `juju integrate` failed with "ERROR connection is shut down"
- https://github.com/canonical/mysql-operators/actions/runs/22079079841/job/63813131367?pr=89 `juju remove-unit` failed with "ERROR connection is shut down"

I'm okay removing the `retry_if_cli_error` because we're only using it in 1 test but, until Juju 3.6.15 is released with https://github.com/juju/juju/pull/21770, they will likely keep failing.

astrojuanlu (dismissed):
Other than that, LGTM!

astrojuanlu:
Also should we bring this to 8.4 as well?

sinclert-canonical:
> Notice that the changes are not equivalent. juju.wait does have a delay parameter, but every CLI invokation is susceptible of failing.
> ...
> I'm okay removing the retry_if_cli_error because we're only using it in 1 test but, until Juju 3.6.15 is released with https://github.com/juju/juju/pull/21770, they will likely keep failing.

I think you are right: every single Jubilant model invocation is susceptible for a _connection closed_ error, however, it is way more prominent in tests that perform a lot of quick invocations (some of them in a loop).

IIRC, this was a regression introduced in Juju 3.6.13, which was way less noticeable in its previous patch version. If changes in PR https://github.com/juju/juju/pull/21770 are included in 3.6.15, we will not need to be so precautious all over the place.

I think we should decouple the fixes, so:
- We fix the Juju spaces 2.9 test ASAP (critical fix pointed out in [this comment](https://github.com/canonical/mysql-operators/pull/89#discussion_r2813200089)).
- Leave the rest of the changed in draft mode, including the one getting rid of the `retry_if_cli_error` snowflake. Those will depend on whether the upcoming Juju 3.6.15 reverts the regression or not.

---

Question for @paulomach:

Why to include deployment constraints all over the place? I read in the PR description that Juju was sometimes getting the wrong architecture, but AFAIK this has never been the case. Do you have an example?

astrojuanlu (dismissed):

sinclert-canonical (dismissed):

sinclert-canonical (approved):

astrojuanlu (approved):

paulomach:
> Question for @paulomach:
> 
> Why to include deployment constraints all over the place? I read in the PR description that Juju was sometimes getting the wrong architecture, but AFAIK this has never been the case. Do you have an example?

For history: https://github.com/juju/juju/issues/21822

PR #90, opened 2026-02-16
[DPE-9035] Tweak password validation logic

astrojuanlu:
## Issue

- Changed minimum length to 12
- Added maximum length on the charm side of 130
- Kept default length of automatically generated passwords to 24

## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.

PR #94, opened 2026-02-17
[DPE-7896] Switch to MySQL Router 8.4

sinclert-canonical:
This PR switches the default channel for MySQL Router within the integration tests to target `8.4/edge`.

When making this change, and given the _subordinate nature_ of the MySQL Router VM operator, we also needed to bump the base of the MySQL Test app to use the Ubuntu 24.04 one. So, instead of doing so in a single test, I am doing that across the board.

---

Depends on the publication of MySQL Router to the `8.4/edge` channel.

### DIFF
--- a/kubernetes/src/relations/mysql_provider.py
+++ b/kubernetes/src/relations/mysql_provider.py
@@ -139,6 +139,7 @@ class MySQLProvider(Object):
         ) as e:
             logger.exception("Failed to set up database relation", exc_info=e)
             self.charm.unit.status = BlockedStatus("Failed to create scoped user")
+            return
 
         try:
             # create k8s services for endpoints
@@ -158,6 +159,7 @@ class MySQLProvider(Object):
                 "Permission to create k8s services denied. `juju trust`"
             )
             event.defer()
+            return
 
         # Set relation data
         self.database.set_endpoints(relation_id, f"{primary_endpoint}:3306")
--- a/machines/src/relations/mysql_provider.py
+++ b/machines/src/relations/mysql_provider.py
@@ -241,6 +241,7 @@ class MySQLProvider(Object):
                 "%",
                 extra_roles=extra_user_roles,
             )
+
             db_version = self.charm._mysql.get_mysql_version()
         except (
             MySQLCreateApplicationDatabaseError,
@@ -249,6 +250,7 @@ class MySQLProvider(Object):
         ) as e:
             logger.exception("Failed to set up database relation", exc_info=e)
             self.charm.unit.status = BlockedStatus("Failed to set up relation")
+            return
 
         rw_endpoints, ro_endpoints, _ = self.charm.get_cluster_endpoints(DB_RELATION_NAME)
         self.database.set_database(relation_id, db_name)
--- a/kubernetes/tests/integration/integration/high_availability/test_async_replication.py
+++ b/kubernetes/tests/integration/integration/high_availability/test_async_replication.py
@@ -141,8 +141,8 @@ def test_deploy_router_and_app(first_model: str) -> None:
     model_1.deploy(
         charm=MYSQL_ROUTER_NAME,
         app=MYSQL_ROUTER_NAME,
-        base="ubuntu@22.04",
-        channel="8.0/edge",
+        base="ubuntu@24.04",
+        channel="8.4/edge",
         num_units=1,
         trust=True,
     )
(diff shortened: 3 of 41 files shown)


kubernetes/src/relations/mysql_provider.py
```diff
@@ -139,6 +139,8 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
         ) as e:
             logger.exception("Failed to set up database relation", exc_info=e)
             self.charm.unit.status = BlockedStatus("Failed to create scoped user")
+            event.defer()
```
astrojuanlu: I think this `event.defer()` wasn't there in the old version 🤔 

https://github.com/canonical/mysql-operators/blob/fc2159a6c52eee2d37e6e55aa831f6e389caf0ef/kubernetes/src/relations/mysql_provider.py#L158-L161

machines/src/relations/mysql_provider.py
```diff
@@ -249,6 +250,8 @@ def _on_database_requested(self, event: DatabaseRequestedEvent):
         ) as e:
             logger.exception("Failed to set up database relation", exc_info=e)
             self.charm.unit.status = BlockedStatus("Failed to set up relation")
+            event.defer()
```
astrojuanlu: Same as above

https://github.com/canonical/mysql-operators/blob/fc2159a6c52eee2d37e6e55aa831f6e389caf0ef/machines/src/relations/mysql_provider.py#L258-L262

kubernetes/src/relations/mysql_provider.py
sinclert-canonical: I know, but we were also missing that as part of the error handling.

This method is executed whenever a user relates to the MySQL Server charm using the `mysql_client` interface. Whenever the requested database / scoped user cannot be created (i.e. workload unreachable / bad state), we should not make the user to unrelate + relate again, but to retry in the event cycle.

astrojuanlu: But then, wouldn't we be misusing the "Blocked" status? https://documentation.ubuntu.com/juju/3.6/reference/status/#workload-charm-status

```
blocked  The charm is stuck. Human intervention is required.
```

It seems to me that Blocked should be "final", but maybe I'm misunderstanding the semantics here.

sinclert-canonical: You may be right. There are not many `BlockedStatus` settings prior to an `event.defer()` call.

Will stick to the return.

astrojuanlu:
As @sinclert-canonical noticed, `test_async_replication.py` is failing https://github.com/canonical/mysql-operators/actions/runs/22099344059/job/63878271082#step:12:1533 with

```
Error: Error creating user account: Error creating MySQL account for router (CREATE USER stage): Error executing MySQL query "CREATE USER IF NOT EXISTS 'mysql_router1_e70p0i9'@'%' IDENTIFIED WITH `caching_sha2_password` BY ***": Your password does not satisfy the current policy requirements (1819)
```

This probably has nothing to do with MySQL Router 8.4 tough. Rather, we accidentally broke this test in #67 and the real failure got masked. Commit 994230d714 fixed this, and now the real failure is seen.

Turns out that `mysqlrouter --bootstrap` as called here https://github.com/canonical/mysql-router-operators/blob/cc65b998668405c803a8fe65feb1dc1451d01d52/common/common/workload.py#L283-L287 is generating random passwords with a length fixed to 16:

```
$ tail -f /var/log/mysql/general.log
...
2026-02-17T17:43:39.915622Z       339 Query     CREATE USER IF NOT EXISTS 'mysql_router1_943xq3g'@'%' IDENTIFIED WITH `caching_sha2_password` BY 'M5KLv}RPFUTlDb0f'
2026-02-17T17:43:39.922305Z       339 Query     ROLLBACK
2026-02-17T17:43:39.922759Z       339 Query     CREATE USER IF NOT EXISTS 'mysql_router1_943xq3g'@'%' IDENTIFIED WITH `caching_sha2_password` BY '4Sf4mofQ53[.,Haf'
2026-02-17T17:43:39.924344Z       339 Query     ROLLBACK
2026-02-17T17:43:39.924589Z       339 Query     CREATE USER IF NOT EXISTS 'mysql_router1_943xq3g'@'%' IDENTIFIED WITH `caching_sha2_password` BY 'q,/@5,$03c0AJn48'
2026-02-17T17:43:39.925711Z       339 Query     ROLLBACK
...
```

This length is hardcoded in the source code so there's no way to change it (without recompiling):

https://github.com/mysql/mysql-server/blob/447eb26e/router/src/bootstrap/src/bootstrap_mysql_account.cc#L38

However, https://github.com/canonical/mysql-operators/pull/90/ lowers the minimum length to 12, so we can merge that first and then rebase this on top of that one.

astrojuanlu:
Proof: rebased this PR on top of #90, 4/4 `test_async_replication.py` are passing

astrojuanlu (approved):

paulomach (approved):

PR #98, opened 2026-02-18
[DPE-8295] Refresh V3

sinclert-canonical:
This PR implements refresh V3 support to both the Kubernetes and VM charms.

### Key take-aways
1. Refresh V3 _vanilla implementation_ is less permissive about Kubernetes charms not being trusted. Therefore, all integration tests that were missing the trust flags when deployment are now including it.
2. Refresh V3 leverages `httpx` internally, and the version of `lightkube` that we were using (0.15.0), is incompatible with it. Bumping it requires a newer version of Rust, which I bumped as well.
3. Extending the `_on_config_change` event handler raised a linting error about the length of such method. Instead of ignoring it, I moved all the workload specific config into a new `_config_change_update` method.

### Disclaimers:
- Upgrade tests are expected to fail: refresh V3 charms can only be refreshed from a previous refresh V3 charm.
- The hard-coded `charm` key within the refresh versions YAML files needs to be reverted before merging.

### References:

- Charm-refresh [documentation](https://canonical-charm-refresh.readthedocs-hosted.com/latest/).


kubernetes/tests/integration/integration/high_availability/test_replication_logs_rotation.py
```diff
@@ -49,6 +50,7 @@ def test_deploy_highly_available_cluster(juju: Juju, charm: str) -> None:
         channel="latest/edge",
         config={"sleep_interval": 300},
         num_units=1,
+        trust=False,
```
taurus-forever, line 53: Q: Isn't it a default one? LGTM.

machines/src/refresh.py
```diff
@@ -0,0 +1,125 @@
...
+        self,
+        *,
+        snap_name: str,
+        snap_revision: str,
+        refresh: charm_refresh.Machines,
+    ) -> None:
+        """Refreshes the installed snap."""
+        # TODO add graceful shutdown before refreshing snap?
+        # TODO: Future improvement
+        # If snap refresh fails (i.e. same snap revision installed) after graceful shutdown, restart workload
```
taurus-forever, line 122: Should be moved to backlog as tickets?

machines/refresh_versions.toml
```diff
@@ -0,0 +1,15 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+charm_major = 1
+workload = "8.4"
+
+charm = "8.4/1.0.0" # TODO remove
```
taurus-forever: remove?

taurus-forever (commented):
This is really massive. Great job done!

kubernetes/tests/integration/integration/high_availability/test_replication_logs_rotation.py
sinclert-canonical: Yes. I am adding it so that every deployment in the same method have the same exact arguments.

machines/refresh_versions.toml
sinclert-canonical: Sadly, this cannot be remove while the PR is still getting reviewed / I might iterate on it.

Removing this line would make `charmcraft` not able to build the charm, given the new requirement of _having some kind of version_ in the newly introduced `refresh_versions.toml`. Not ideal by any means, but this is one of the requirements for having refresh V3 Introduced into an existing charm.

kubernetes/tests/integration/integration/test_charm.py
sinclert-canonical, line 1: There is some delay between when the instances are removed due to scaling down, and the cluster information update. Therefore the substitution of _online / offline count_ by just _number of units_.

Do you know why this became apparent with refresh V3? Is there anything special that the charm is doing on scale down? (see proposed `UnitTeardown` juju event [interception](https://github.com/canonical/mysql-operators/blob/51440ec6e4642b2af978ecb8d1f4c3a1ee43c019/kubernetes/src/charm.py#L182-L184)) @paulomach

paulomach (commented):
Before finishing, from manual testing looks good - will get back asap

sinclert-canonical:
🗞️ In order to ease the review process, this PR has been split into 5 different PRs:

- https://github.com/canonical/mysql-operators/pull/147
- https://github.com/canonical/mysql-operators/pull/148
- https://github.com/canonical/mysql-operators/pull/149
- https://github.com/canonical/mysql-operators/pull/150

PR #99, opened 2026-02-19
[DPE-9035] Validate passwords (8.0/edge)

astrojuanlu:
## Issue

Backports #67 + #90 to `8.0/edge`.

## Checklist


astrojuanlu:
The CI is definitely not in a good state. Lots of controllers failed to bootstrap, timeout errors... I'll re-trigger when #89 lands.

machines/src/relations/mysql_provider.py
```diff
@@ -235,14 +235,6 @@ def _on_database_requested(self, event: DatabaseRequestedEvent):
         self.charm.update_endpoint_address(DB_RELATION_NAME)
 
         try:
-            db_version = self.charm._mysql.get_mysql_version()
-            rw_endpoints, ro_endpoints, _ = self.charm.get_cluster_endpoints(DB_RELATION_NAME)
-            self.database.set_database(relation_id, db_name)
-            self.database.set_credentials(relation_id, db_user, db_pass)
```
paulomach, line 241: Besides the obvious `password = generate_random_password(DEFAULT_PASSWORD_LENGTH)` 
how these changes apply to the password validation?

kubernetes/src/relations/mysql_provider.py
```diff
@@ -120,21 +120,10 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
...
-            # create k8s services for endpoints
-            self.charm.k8s_helpers.create_endpoint_services(["primary", "replicas"])
-
-            primary_endpoint = dotappend(socket.getfqdn(f"{self.charm.app.name}-primary"))
-            replicas_endpoint = dotappend(socket.getfqdn(f"{self.charm.app.name}-replicas"))
-
-            db_version = self.charm._mysql.get_mysql_version()
-
-            # wait for endpoints to be ready
-            self.charm.k8s_helpers.wait_service_ready((primary_endpoint, 3306))
```
paulomach, line 136: Same as https://github.com/canonical/mysql-operators/pull/99/changes#r2842680128

paulomach (commented):
Looks good, but I did not understood the changes on the `mysql_provider.py`s

kubernetes/src/relations/mysql_provider.py
```diff
@@ -149,22 +138,24 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
...
-            logger.info(f"Created user for app {app_name}")
-            self.charm.unit.status = ActiveStatus()
+            db_version = self.charm._mysql.get_mysql_version()
         except (
             MySQLCreateApplicationDatabaseError,
             MySQLCreateApplicationScopedUserError,
             MySQLGetMySQLVersionError,
         ) as e:
             logger.exception("Failed to set up database relation", exc_info=e)
             self.charm.unit.status = BlockedStatus("Failed to create scoped user")
```
sinclert-canonical, line 148: Remember the missing `return` we discussed in the `8.4/edge` branch.

kubernetes/src/relations/mysql_provider.py
```diff
@@ -175,6 +166,16 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
             )
             event.defer()
```
sinclert-canonical, line 167: Remember the missing `return` we discussed in the `8.4/edge` branch.

kubernetes/tests/integration/conftest.py
```diff
@@ -10,6 +10,19 @@
...
+def pytest_configure(config):
+    """Configure pytest logging format with timestamps."""
+    config.option.log_cli_format = (
+        "%(asctime)s %(levelname)-8s %(name)s:%(filename)s:%(lineno)d %(message)s"
+    )
+    config.option.log_cli_date_format = "%b %d %H:%M:%S"
+    config.option.log_format = (
+        "%(asctime)s %(levelname)-8s %(name)s:%(filename)s:%(lineno)d %(message)s"
+    )
+    config.option.log_date_format = "%b %d %H:%M:%S"
```
sinclert-canonical, line 23: This snippet slipped through the review process in the `8.4/edge` PR. Either we define it for both VM and K8s integration tests, or we do not define it at all (my preference).

machines/src/relations/mysql_provider.py
```diff
@@ -266,6 +257,16 @@ def _on_database_requested(self, event: DatabaseRequestedEvent):
             logger.exception("Failed to set up database relation", exc_info=e)
             self.charm.unit.status = BlockedStatus("Failed to set up relation")
```
sinclert-canonical, line 258: Remember the missing `return` we discussed in the `8.4/edge` branch.

sinclert-canonical (commented):
LGTM aside of the comments.

I am still worried that we are introducing breaking changes in the MySQL 8.0 charms. I know the password-validation component is only applied to **new** passwords, but we cannot anticipate how the users are going to relate charm (maybe they use one that provides a < 12 length password?), or whether there are some upgrade connotations.

Please, do not just rely on our integration tests. Test some real-world scenarios locally, to make sure we are not breaking anything.

machines/src/relations/mysql_provider.py
sinclert-canonical: Let me provide some context here:

These modifications come from a discussion Juanlu and I had, where we wanted to validate the user passwords _as early as possible_ within the mysql-provider endpoint, so we can avoid creating a dedicated `validate_user_password` function in our MySQL lib.

To that end, we proposed to reshuffled the code ordering so that:
1. Workload actions are executed first (i.e. create database, create user, fetch version...).
2. Networking actions are executed next (i.e. create K8s endpoints...)
3. Juju databag actions are executed last (i.e. set credentials, set version...)

kubernetes/tests/integration/conftest.py
paulomach: I share the preference

machines/src/relations/mysql_provider.py
paulomach: Thanks for the clarification, no issues

paulomach:
@astrojuanlu , on upgrade how does the component gets installed? I don't think I saw that. Ping me on how to test it - I agree with @sinclert-canonical current CI cases are not enough

astrojuanlu:
I think this is the key bit:

> I am still worried that we are introducing breaking changes in the MySQL 8.0 charms. [...] we cannot anticipate how the users are going to relate charm (maybe they use one that provides a < 12 length password?),

I agree that, essentially, this is a breaking change. If we all agree we want to move forward despite this fact, I'm happy to do more manual testing, and even codify some real-world scenarios into new integration tests. But no matter how well we test, we're indeed changing the behaviour and this can have unintended consequences.

sinclert-canonical:
Should we close this?

astrojuanlu:
In the end, we decided to not move forward with adding this breaking change to the 8.0/edge branch. Password validation rules will only be enforced in the 8.4 releases. Closing.

PR #102, opened 2026-02-19
[MISC] Integration test fixes

astrojuanlu:
## Issue

Backports @paulomach 's #89 to `8.4/edge`

I included the fix for `test_spaced_db.py`, happy to split to its own PR if needed

## Checklist

### DIFF
--- a/kubernetes/tests/integration/integration/high_availability/test_async_replication.py
+++ b/kubernetes/tests/integration/integration/high_availability/test_async_replication.py
@@ -84,6 +84,7 @@ def test_build_and_deploy(first_model: str, second_model: str, charm: str) -> No
         constraints=constraints,
         resources=resources,
         num_units=3,
+        trust=True,
     )
     model_2 = Juju(model=second_model)
     model_2.deploy(
@@ -94,6 +95,7 @@ def test_build_and_deploy(first_model: str, second_model: str, charm: str) -> No
         constraints=constraints,
         resources=resources,
         num_units=3,
+        trust=True,
     )
 
     logging.info("Waiting for the applications to settle")
@@ -137,6 +139,7 @@ def test_async_relate(first_model: str, second_model: str) -> None:
 def test_deploy_router_and_app(first_model: str) -> None:
     """Deploy the router and the test application."""
     logging.info("Deploying the router and test application")
+    constraints = {"arch": architecture.architecture}
     model_1 = Juju(model=first_model)
     model_1.deploy(
         charm=MYSQL_ROUTER_NAME,
@@ -145,6 +148,7 @@ def test_deploy_router_and_app(first_model: str) -> None:
         channel="8.4/edge",
         num_units=1,
         trust=True,
+        constraints=constraints,
     )
     model_1.deploy(
         charm=MYSQL_TEST_APP_NAME,
@@ -153,6 +157,7 @@ def test_deploy_router_and_app(first_model: str) -> None:
         channel="latest/edge",
         num_units=1,
         trust=False,
+        constraints=constraints,
     )
 
     logging.info("Relating the router and test application")
--- a/kubernetes/tests/integration/integration/high_availability/test_async_replication_upgrade.py
+++ b/kubernetes/tests/integration/integration/high_availability/test_async_replication_upgrade.py
@@ -86,6 +86,7 @@ def test_build_and_deploy(first_model: str, second_model: str, charm: str) -> No
         constraints=constraints,
         resources=resources,
         num_units=3,
+        trust=True,
     )
     model_2 = Juju(model=second_model)
     model_2.deploy(
@@ -96,6 +97,7 @@ def test_build_and_deploy(first_model: str, second_model: str, charm: str) -> No
         constraints=constraints,
         resources=resources,
         num_units=3,
+        trust=True,
     )
 
     logging.info("Waiting for the applications to settle")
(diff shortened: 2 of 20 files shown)


paulomach:
See #103

astrojuanlu:
The tests failures here are
- `connection is shut down` 5x,
- a leftover I'm about to fix,
- various timeouts,

but none of them is `filesystem for storage instance "database/X" not found`. Which checks out, because #103 in principle is for Juju 2.9 issues, which isn't supported in 8.4/edge.

So my understanding is that I don't have to backport that. Do you agree?

paulomach:
> The tests failures here are
> 
>     * `connection is shut down` 5x,
> 
>     * a leftover I'm about to fix,
> 
>     * various timeouts,
> 
> 
> but none of them is `filesystem for storage instance "database/X" not found`. Which checks out, because #103 in principle is for Juju 2.9 issues, which isn't supported in 8.4/edge.
> 
> So my understanding is that I don't have to backport that. Do you agree?

Indeed, forgot about this detail!

Btw, don't work on the weekend !!

paulomach (approved):

sinclert-canonical (commented):
> I included the fix for test_spaced_db.py, happy to split to its own PR if needed

I would appreciate that, yes.

I think you folks know what I think about these changes: they are just work-arounds, as the real instability came with Juju 3.6.14. Therefore, given that we do not have any rush to publish the MySQL 8.4 charm (contrary to MySQL 8.0), I would say to delay these changes until we see how our test suite behaves whenever Juju 3.6.15 is released.

paulomach:
> > I included the fix for test_spaced_db.py, happy to split to its own PR if needed
> 
> I would appreciate that, yes.
> 
> I think you folks know what I think about these changes: they are just work-arounds, as the real instability came with Juju 3.6.14. Therefore, given that we do not have any rush to publish the MySQL 8.4 charm (contrary to MySQL 8.0), I would say to delay these changes until we see how our test suite behaves whenever Juju 3.6.15 is released.

Though I don't fully agree - as we mimic a user experience when using juju, e.g. if a status call fails, the user just re run it - I am ok with delaying it until 3.6.15 is out to validate the fixes. If issues still there, we should merge.
The goal of the integration tests is to assure charm functionality, not tooling limitations. Workarounds for the tooling are ok, as long as we do our part of properly reporting and tracking, as @astrojuanlu is gracefully doing.

astrojuanlu:
Will revisit after we test with the upcoming Juju version, which includes a fix for https://github.com/juju/juju/issues/21664

astrojuanlu:
Looking at the last scheduled run of the 8.4/edge branch (already including Juju 3.6.18+), https://github.com/canonical/mysql-operators/actions/runs/23031919468 , the majority of the tests touched here are no longer failing, and we already identified that async replication is broken in the upcoming version. So we don't think it's necessary to proceed with many of these fixes.

However, I do think it would be good to keep these from #89:

> - add missing trust on k8s - though not always relevant, keeps it consistent
> - add architecture constraints to router/test-app - when missing, juju might select wrong revision/arch on k8s - resulting in test timeout due to app pod(s) never being scheduled, as it will have a architecture bound node affinity rule

So I've kept that commit. If you folks still disagree, I'll proceed and close.

astrojuanlu:
And async replication tests fixed themselves! The new 3.6/candidate clearly got back on track. Only 1 failed because of an unrelated issue.

I'll restart the CI, but I think this is now ready to go in @paulomach @sinclert-canonical .

paulomach:
> And async replication tests fixed themselves! The new 3.6/candidate clearly got back on track. Only 1 failed because of an unrelated issue.
> 
> I'll restart the CI, but I think this is now ready to go in @paulomach @sinclert-canonical .

I'd agree.

sinclert-canonical:
> I'll restart the CI, but I think this is now ready to go in @paulomach @sinclert-canonical.

I am happy with adding the `trust` flag to every MySQL K8s deployment (something I am also proposing in the Refresh V3 PR IV), but I am still confused about the architecture thing. Have you tried the CI without it? Do we still see the wrong MySQL Test App being deployed using juju `3.6.19`?

astrojuanlu:
> but I am still confused about the architecture thing. Have you tried the CI without it?

I have not

> Do we still see the wrong MySQL Test App being deployed using juju 3.6.19?

from @paulomach 's original comment on #89, 

> when missing, juju might select wrong revision/arch on k8s - resulting in test timeout due to app pod(s) never being scheduled, as it will have a architecture bound node affinity rule

I don't know how easy to spot are these timeouts, do you remember past examples @paulomach ? Happy to remove the arch constraints as well but I'm concerned that I might get false positives, or that the timeouts come back in the future.

paulomach:
> > but I am still confused about the architecture thing. Have you tried the CI without it?
> 
> I have not
> 
> > Do we still see the wrong MySQL Test App being deployed using juju 3.6.19?
> 
> from @paulomach 's original comment on #89,
> 
> > when missing, juju might select wrong revision/arch on k8s - resulting in test timeout due to app pod(s) never being scheduled, as it will have a architecture bound node affinity rule
> 
> I don't know how easy to spot are these timeouts, do you remember past examples @paulomach ? Happy to remove the arch constraints as well but I'm concerned that I might get false positives, or that the timeouts come back in the future.

I keep seeing this when running tests on arm machines. Cannot be sure on CI since we don't have log from k8s (like pod describe)

astrojuanlu:
With that in mind, @sinclert-canonical do you agree with keeping the arch constraints then?

sinclert-canonical (approved):

PR #105, opened 2026-02-23
bugfix: Use tls-ca-chain received from S3 relation properly in all S3 connections in the charm

### LINKED ISSUE #92: [k8s] [vm] mysql charms do not work with S3 clouds with custom CA
## Steps to reproduce

1. Prepare a S3 cloud, with custom CA certificate. For an example, microceph radosgw can be considered.
(follow the steps as per the test_backups_ceph integration test).
2. Deploy and configure S3 integrator with custom CA certificate
3. Deploy mysql charm (either vm or k8s -- the problem exists in both).
4. Integrate mysql charm with S3 integrator.
5. Try actions like `create-backup` and `list-backups`.

## Expected behavior
The charm integration works, and the backup actions `create-backup` and `list-backups` should work.

## Actual behavior

The charm fails to list / create backup, because it fails to use the custom TLS certificate in the right way. 
You can see errors similar to the following in juju logs:

```
'utf-8' codec can't decode byte 0x88 in position 2: invalid start byte
```

If you look at the following piece of code:

```python
    with tempfile.NamedTemporaryFile() if ca_chain else nullcontext() as ca_file:
        if ca_file:
            ca = "\n".join([base64.b64decode(s).decode() for s in ca_chain])
            ca_file.write(ca.encode())
            ca_file.flush()

        s3 = session.resource(
            "s3",
            endpoint_url=_construct_endpoint(s3_parameters),
            verify=ca_file.name if ca_file else True,
        )
```

in s3_helpers.py, you can see that the charm tries to b64decode the value that it gets as ca_chain from `s3-integrator`. However, the value is already a base64 decoded string, ready for consumption. This extra decoding of already decoded string will thus fail, because then invalid characters will be produced after decoding.

Additional Observation:
It seems that we don't have tests in our integration tests for this particular behavior where local S3 provider using custom CA certificate is used. 

## Versions

Operating system: noble 24.04

Juju CLI: 3.6

Charm revision: Any revision from 8.0/stable

theoctober19th:
## Issue
Fixes #92 

## Solution
* Do not decode `tls-ca-chain` received via S3 relation since it is already decoded.
* Use the `tls-ca-chain` received via S3 relation in all places where S3 connection is made
* Use TLS certificates with S3 in the Ceph Backup integration tests.

## Checklist


sinclert-canonical:
👋🏻 Thanks for addressing this issue! Is this ready to get reviewed?

theoctober19th:
Hi @sinclert-canonical -- yes, I believe the PR is ready for review.

paulomach (approved):
Thank you @theoctober19th 🚀

sinclert-canonical (approved):

taurus-forever:
I believe you do not need my review here, as you have 2 pair of eyes already.

However, this scary me as it looks like backward incompatible.

Was it ever worked? Was ca_chain decoded before? Will this brake https://github.com/canonical/mysql-operator/issues/395 CC: @sabaini (any comments on decode part there)?  Is it because users normally encode s3-integrator config https://charmhub.io/s3-integrator/configurations#tls-ca-chain  ?

Tnx!

theoctober19th:
Hi reviewers (cc @sinclert-canonical @paulomach ), the original patch was not sufficient, because I noticed that we don't use the CA certificate provided via the S3 relation when we do list backups action. This was missed completely by CI because we have no tests in CI that uses custom TLS certificate for S3.

Therefore, I'm editing the PR to add both the fix in the list  backups action, and also the proper tests in the integration tests to validate that.

theoctober19th:
Hi @paulomach and @sinclert-canonical. I've udpated the PR with the usage of custom TLS for `xtrabackup` CLI calls, and also updated the integration tests using ceph to actually use TLS certificates.

However I also see S3 connection being made `/start-mysql-pitr-helper-collector.sh`, perhaps the changes are also needed at rock / snap level. I lack the context behind the charms and rocks a little bit, and the scope of changes is increasing. Can you please help me take this forward from here?

theoctober19th:
After our offline sync (me and @sinclert-canonical) -- I'm going to split this PR into two:
* https://github.com/canonical/mysql-operators/pull/156 (another PR) that attempts to fix just the dual-decoding problem I initially reported in #92  
* #105 (this PR) that attempts to add S3 TLS across all different places where S3 used in the mysql charm (eg, `xbcloud` calls, etc.) and also add integration tests in Ceph tests so that the S3 TLS behavior doesn't remain untested in the MySQL charm. I chose to keep the current changes here, because I use the test releases from this branch in [S3 Integrator CI](https://github.com/canonical/object-storage-integrator/actions/runs/23042905035/job/66926830393), and any change I make to the fixes in this branch will break that CI. 

It is to be noted however that just merging #156 won't solve the issue of being unable to use custom TLS in MySQL for us, since with just the dual-decoding behavior fixed, the `list-backups` action will work but the `restore` would not work in custom S3 TLS scenario (because `xblcloud` does not yet use custom TLS in the charm). Therefore, when we finally merge the PR, I do still advocate for merging this and the other PR #156 together.

In addition, I see that the S3 parameters are also used in `/start-mysql-pitr-helper-collector.sh`, and thus with that the scope of this PR has increased (probably needs change in snap / rock level). If any further change in other components are required, I believe it would be best for me to handover this PR to the team.

cc: @paulomach

sinclert-canonical:
> In addition, I see that the S3 parameters are also used in /start-mysql-pitr-helper-collector.sh, and thus with that the scope of this PR has increased (probably needs change in snap / rock level). If any further change in other components are required, I believe it would be best for me to handover this PR to the team.

That is all fine by me (recommended even, given the regrettable state of the PITR helper).

Is this ready to review then?

theoctober19th:
@sinclert-canonical  Yes -- in that case, the PR is ready for review.

.github/workflows/ci.yaml
```diff
@@ -117,3 +117,25 @@ jobs:
       artifact-prefix: ${{ needs.build.outputs.artifact-prefix }}
       path-to-charm-directory: ${{ matrix.path }}
     secrets: inherit
+
+  release-test-branch:
```
sinclert-canonical: This whole step needs to be reverted before merging, as previously discussed.

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2435,6 +2436,30 @@ def get_available_memory(self) -> int:
         """Platform dependent method to get the available memory for mysql-server."""
         raise NotImplementedError
 
+    def crete_ca_pem_file(
```
sinclert-canonical:
```diff
-    def crete_ca_pem_file(
+    def create_ca_pem_file(
```

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2491,9 +2523,14 @@ def execute_backup_commands(
...
-            f"{s3_path}",
         ]
-
+        if ca_chain and ca_file_location:
+            xtrabackup_commands += [
+                f"--cacert={ca_file_location}",
+            ]
+        xtrabackup_commands += [
+            f"{s3_path}"
+        ]
```
sinclert-canonical:
```diff
-        if ca_chain and ca_file_location:
-            xtrabackup_commands += [
-                f"--cacert={ca_file_location}",
-            ]
-        xtrabackup_commands += [
-            f"{s3_path}"
-        ]
+       xtrabackup_commands.append(f"--cacert={ca_file_location}") if ca_chain and ca_file_location else None
+       xtrabackup_commands.append(f"{s3_path}")
```

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2546,6 +2583,7 @@ def delete_temp_backup_directory(
             logger.error("Failed to delete temp backup directory")
             raise MySQLDeleteTempBackupDirectoryError from e
 
+
```
sinclert-canonical: Remove white-space

kubernetes/lib/charms/mysql/v0/s3_helpers.py
```diff
@@ -277,25 +319,29 @@ def fetch_and_check_existence_of_s3_path(path: str, s3_parameters: dict[str, str
...
+            aws_secret_access_key=s3_parameters["secret-key"],
+            endpoint_url=_construct_endpoint(s3_parameters),
+            region_name=s3_parameters["region"] or None,
+            verify=ca_file_name if ca_file_name else True,
         )
-        raise
+        try:
+            response = s3_client.get_object(
+                Bucket=s3_parameters["bucket"], Key=path, Range="0-1"
+            )
```
sinclert-canonical: Nit: either place all arguments in the same line (if line length < 100), or one per line using trailing commas.

kubernetes/tests/integration/integration/test_backup_ceph.py
```diff
@@ -60,16 +63,44 @@
...
+    with local_tmp_folder("temp-certs") as tmp_folder:
+        yield tmp_folder
+
+
+@pytest.fixture(scope="session")
+def host_ip() -> str:
+    """The IP address of the host running these tests."""
+    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+        s.connect(("1.1.1.1", 80))
+        return s.getsockname()[0]
```
sinclert-canonical, line 99: What is wrong with the previous implementation?

kubernetes/tests/integration/integration/test_backup_ceph.py
```diff
@@ -78,12 +109,101 @@ def microceph():
             os.environ["CEPH_SECRET_KEY"],
             MICROCEPH_BUCKET,
         )
+    logger.info("Setting up TLS certificates")
+    subprocess.run(["openssl", "genrsa", "-out", str(certs_path / "ca.key"), "2048"], check=True)
+    subprocess.run(
```
sinclert-canonical, line 117: With this long list of arguments, is hard to visualize how the resulting command would look like.

Could we define it as a string that we expand using split()?

kubernetes/tests/integration/integration/test_backup_ceph.py
```diff
@@ -78,12 +109,101 @@ def microceph():
...
+            "-subj",
+            f"/C=US/ST=Denial/L=Springfield/O=Dis/CN={host_ip}",
+        ],
+        check=True,
+    )
+    subprocess.run(
+        ["openssl", "genrsa", "-out", str(certs_path / "server.key"), "2048"],
+        check=True,
+    )
+    subprocess.run(
```
sinclert-canonical, line 127: With this long list of arguments, is hard to visualize how the resulting command would look like.

Could we define it as a string that we expand using split()?

kubernetes/tests/integration/integration/test_backup_ceph.py
```diff
@@ -78,12 +109,101 @@ def microceph():
...
+            "-subj",
+            f"/C=US/ST=Denial/L=Springfield/O=Dis/CN={host_ip}",
+        ],
+        check=True,
+    )
+
+    with open(certs_path / "extfile.cnf", "w") as extfile:
+        extfile.write(f"subjectAltName = DNS:{host_ip}, IP:{host_ip}")
 
+    subprocess.run(
```
sinclert-canonical, line 137: With this long list of arguments, is hard to visualize how the resulting command would look like.

Could we define it as a string that we expand using split()?

machines/tests/unit/test_mysql.py
```diff
@@ -1192,6 +1217,20 @@ def test_retrieve_backup_with_xbcloud(
             "--directory",
             "mysql/data/directory/#mysql_sst_XXXX",
         ]
+        _expected_ca_dir_commands = [
+            "mktemp",
+            "--directory",
+            "mysql/data/directory/s3_ca_XXXX",
+        ]
+        _exepected_ca_content_commands = [
```
sinclert-canonical, line 1225: With this long list of arguments, is hard to visualize how the resulting command would look like.

Could we define it as a string that we expand using split()?

machines/tests/unit/test_mysql.py
```diff
@@ -1013,14 +1016,28 @@ def test_execute_backup_commands(self, _execute_commands):
...
             "mktemp",
             "--directory",
             "/tmp/base/directory/xtra_backup_XXXX",
         ]
+        _expected_ca_dir_commands = [
+            "mktemp",
+            "--directory",
+            "/tmp/base/directory/s3_ca_XXXX",
+        ]
+        _exepected_ca_content_commands = [
```
sinclert-canonical, line 1032: With this long list of arguments, is hard to visualize how the resulting command would look like.

Could we define it as a string that we expand using split()?

sinclert-canonical (commented):
🧡 Thanks again for helping us get the backup functionality into a better state.

I have provided a 1st round of review, with most of the comments being readability / style wise, as I assume you know what you are doing. Please, bear in mind that the same exact comments I left for the K8s substrate, **are applicable to VM as well**.

.github/workflows/ci.yaml
theoctober19th: This is noted, will be removed before the merge. Thanks for the reminder.

kubernetes/tests/integration/integration/test_backup_ceph.py
theoctober19th: The previous implementation resolved to loopback interface, and it wasn't working when I run the tests locally. However, this one resolves to the private IP address from different interface, and is both runnable locally as well as in CI.

sinclert-canonical: Ok, that is enough for me to go ahead. Please Paulo / Juanlu, tune in if you see any problems with Cloudfare DNS resolver (i.e. `1.1.1.1`)

machines/lib/charms/mysql/v0/mysql.py
```diff
@@ -2493,9 +2523,11 @@ def execute_backup_commands(
             f"--s3-endpoint={s3_parameters['endpoint']}",
             f"--s3-api-version={s3_parameters['s3-api-version']}",
             f"--s3-bucket-lookup={s3_parameters['s3-uri-style']}",
-            f"{s3_path}",
         ]
-
+        xtrabackup_commands.append(
+            f"--cacert={ca_file_location}"
+        ) if ca_chain and ca_file_location else None
```
sinclert-canonical, line 2559: Do we have different line length requirements in the `machines` charm? 🤔

machines/lib/charms/mysql/v0/mysql.py
```diff
@@ -2587,6 +2623,11 @@ def retrieve_backup_with_xbcloud(
             f"--s3-bucket-lookup={s3_parameters['s3-uri-style']}",
             f"--s3-api-version={s3_parameters['s3-api-version']}",
             f"{s3_parameters['path']}/{backup_id}",
+        ]
+        retrieve_backup_command.append(
+            f"--cacert={ca_file_location}"
+        ) if ca_chain and ca_file_location else None
```
sinclert-canonical, line 2659: Do we have different line length requirements in the `machines` charm? 🤔

sinclert-canonical (approved):
LGTM overall, with a couple of comments.

Please wait for Paulo / Juanlu reviews, as they may see things I missed.

machines/lib/charms/mysql/v0/mysql.py
theoctober19th: This is because the `kubernetes` charm does not include `lib_path` when running formatters. I think it's better to handle the formatting of lib in a future PR, since running formatter on the lib now produces a lot of unrelated diff.

machines/lib/charms/mysql/v0/mysql.py
theoctober19th: This is because the `kubernetes` charm does not include `lib_path` when running formatters. I think it's better to handle the formatting of lib in a future PR, since running formatter on the lib now produces a lot of unrelated diff.

paulomach (approved):
Thank you @theoctober19th , we are probably need to rebase from 8.4/edge once #144 to make tests more robust

kubernetes/tests/integration/integration/test_backup_ceph.py
astrojuanlu: It should be quite reliable, no problem from my side.

machines/tests/unit/test_mysql.py
astrojuanlu: IIUC, this feedback was addressed for some of the commands, but not all

astrojuanlu (approved):

astrojuanlu:
> In addition, I see that the S3 parameters are also used in `/start-mysql-pitr-helper-collector.sh`, and thus with that the scope of this PR has increased (probably needs change in snap / rock level). If any further change in other components are required, I believe it would be best for me to handover this PR to the team.

Sounds sensible to me! 👍🏼

machines/tests/unit/test_mysql.py
theoctober19th: My apologies, I think the refactoring was missed in some files. I'll fix them in the next commit.

theoctober19th: @sinclert-canonical @astrojuanlu -- the changes are being reverted by the ruff formatter we're using. To my understanding, this is flagged and fixed by the formatter as substandard. We can always put a `#noqa` at the end of each of these lines -- but I believe it's not worth the effort. What do you think?

astrojuanlu: I actually prefer the list form, but then saw this bit of feedback by @sinclert-canonical . Definitely not a blocker, maybe we can put `# noqa` in some pathological cases but I agree it's not worth it.

PR #146, opened 2026-03-12
[DPE-8295] Refresh V3 (I) - Bump incompatible deps

sinclert-canonical:
**This is the 1st PR from the refresh V3 implementation efforts**

This PR bumps the refresh V3 incompatible dependencies. The version of `lightkube` that we were using (i.e. 0.15.0), is incompatible with the upcoming version of `httpx`. Bumping it requires a newer version of Rust as well.

### DIFF
--- a/kubernetes/charmcraft.yaml
+++ b/kubernetes/charmcraft.yaml
@@ -49,7 +49,7 @@ parts:
       apt-get install rustup -y
 
       rustup set profile minimal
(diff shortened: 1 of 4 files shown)


astrojuanlu (approved):
Hopefully uncontroversial 👍🏼

paulomach (approved):

PR #147, opened 2026-03-12
[DPE-8295] Refresh V3 (II) - Harmonize config parsing

sinclert-canonical:
**This is the 2nd PR from the refresh V3 implementation efforts**

This PR harmonizes the charm configuration parsing, in order to abstract away file / path related logic, which is one of the main differences between K8s and VM substrates.

---

Depends on:

### DIFF
--- a/kubernetes/lib/charms/mysql/v0/mysql.py
+++ b/kubernetes/lib/charms/mysql/v0/mysql.py
@@ -2965,6 +2965,23 @@ class MySQLBase(ABC):
         raise NotImplementedError
 
     @abstractmethod
+    def write_content_to_file(
+        self,
+        path: str,
+        content: str,
+        owner: str,
+        group: str,
+        permission: int,
+    ) -> None:
+        """Write content to file."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def read_file_content(self, path: str) -> str | None:
+        """Reads a file content."""
+        raise NotImplementedError
+
+    @abstractmethod
     def reset_data_dir(self) -> None:
         """Reset the data directory."""
         raise NotImplementedError
--- a/kubernetes/src/charm.py
+++ b/kubernetes/src/charm.py
@@ -614,14 +614,13 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
         if not config_content:
             return
 
-        old_config = self.mysql_config.custom_config(config_content)
-        new_config = self._write_mysqld_configuration()
-
-        # always setup log rotation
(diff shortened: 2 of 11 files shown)


kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2964,6 +2964,23 @@ def wait_until_mysql_connection(self, check_port: bool = True) -> None:
         """
```
astrojuanlu, line 2964: Do we need to bump the lib version?

kubernetes/src/charm.py
```diff
@@ -612,17 +612,16 @@ def _on_config_changed(self, _: EventBase) -> None:
 
         config_content = self._mysql.read_file_content(MYSQLD_CONFIG_FILE)
         if not config_content:
-            # empty config means not initialized, skipping
             return
 
-        previous_config_dict = self.mysql_config.custom_config(config_content)
+        old_config = self.mysql_config.custom_config(config_content)
+        new_config = self._write_mysqld_configuration()
```
astrojuanlu: Does it make any difference to write the mysqld config _before_ log rotation rather than after?

machines/src/config.py
```diff
@@ -43,16 +38,11 @@ def filter_static_keys(self, keys: set) -> set:
...
         return keys - self.static_config
 
-    @property
-    def custom_config(self) -> dict | None:
-        """Return current custom config dict."""
-        if not os.path.exists(self.config_file_path):
-            return None
-
+    @staticmethod
+    def custom_config(config_content: str) -> dict:
```
astrojuanlu: Nit: `get_custom_config`? (So it's clear it's a method)

astrojuanlu (approved):
Approved with some comments

kubernetes/lib/charms/mysql/v0/mysql.py
sinclert-canonical: No, these Python modules does not behave like a charm lib anymore.

kubernetes/src/charm.py
sinclert-canonical: It should not matter, as the written config is MySQL specific, while the log rotation refers to the `logrotate` process (which gets its information from the charm config).

machines/src/config.py
sinclert-canonical: Done in commit https://github.com/canonical/mysql-operators/pull/147/commits/3ecf3a4cb827a96bf52ca360215f560b51dc9f59.

paulomach (approved):

PR #148, opened 2026-03-12
[DPE-8295] Refresh V3 (III) - Implementation

sinclert-canonical:
**This is the 3rd PR from the refresh V3 implementation efforts**

This PR introduces refresh V3 logic into the K8s and VM charms, and it cannot be split.

### Disclaimers:
- Upgrade tests have been disabled, as there is no refresh V3 charm published in the `8.4/edge` channels.
- The hard-coded `charm` key within the refresh versions YAML files needs to be reverted before merging.

---

Depends on:
- https://github.com/canonical/mysql-operators/pull/147


kubernetes/src/refresh.py
```diff
@@ -0,0 +1,108 @@
...
+        try:
+            # Set slow shutdown on all instances
+            for unit in self.app_units:
+                self._charm._mysql.set_dynamic_variable(
+                    instance_address=self._charm.get_unit_address(unit),
+                    variable="innodb_fast_shutdown",
+                    value=0,
+                )
+        except MySQLSetVariableError as e:
+            raise charm_refresh.PrecheckFailed("Failed to set slow shutdown") from e
```
sinclert-canonical, line 103: This whole block has been copied (+adapted) from the existing `upgrade.py` module (see [code](https://github.com/canonical/mysql-operators/blob/ba91866e15a517ca57eb4dded4766c6558be843e/kubernetes/src/upgrade.py#L82-L136)). Please, do not review it as a "new piece of code" but as a mere refactor.

machines/src/refresh.py
```diff
@@ -0,0 +1,125 @@
...
+        try:
+            # Set slow shutdown on all instances
+            for unit in self.app_units:
+                self._charm._mysql.set_dynamic_variable(
+                    instance_address=self._charm.get_unit_address(unit, PEER),
+                    variable="innodb_fast_shutdown",
+                    value=0,
+                )
+        except MySQLSetVariableError as e:
+            raise charm_refresh.PrecheckFailed("Failed to set slow shutdown") from e
```
sinclert-canonical, line 105: This whole block has been copied (+adapted) from the existing `upgrade.py` module (see [code](https://github.com/canonical/mysql-operators/blob/ba91866e15a517ca57eb4dded4766c6558be843e/machines/src/upgrade.py#L106-L143)). Please, do not review it as a "new piece of code" but as a mere refactor.

machines/src/mysql_vm_helpers.py
```diff
@@ -16,6 +15,7 @@
 from typing import override
 
 import jinja2
+import tomli
```
astrojuanlu, line 18: Given that tomllib was introduced in 3.11, maybe we could use that? https://docs.python.org/3.12/library/tomllib.html

machines/tests/integration/integration/test_backup_aws.py
```diff
@@ -303,9 +303,6 @@ def test_restore_on_same_cluster(juju: Juju, cloud_configs_aws) -> None:
         == MOVE_RESTORED_CLUSTER_TO_ANOTHER_S3_REPOSITORY_ERROR
     ), "cluster should migrate to blocked status after restore"
 
-    # scale down the cluster to preserve resources for the following tests
-    scale_app_units(juju, DATABASE_APP_NAME, 0)
```
astrojuanlu, line 307: I get that this maybe wasn't needed. But I'm curious, was it preventing proper application of the refresh logic?

kubernetes/src/charm.py
```diff
@@ -117,6 +118,12 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
     def __init__(self, *args):
         super().__init__(*args)
 
+        # Show logger name (module name) in logs
+        root_logger = logging.getLogger()
+        for handler in root_logger.handlers:
+            if isinstance(handler, ops.log.JujuLogHandler):
+                handler.setFormatter(logging.Formatter("{name}:{message}", style="{"))
```
astrojuanlu, line 124: Should this be at the top of the module, rather than in `__init__.py`?

astrojuanlu (approved):
Approved with a few comments, thank you!

machines/src/mysql_vm_helpers.py
sinclert-canonical: We would need the `tomli_w` package as well eventually (see [code](https://github.com/canonical/mysql-operators/blob/sinclert/8295/refresh-v3-part-4/machines/tests/integration/integration/high_availability/test_upgrade.py#L12) for the follow-up PR), and that was never included into Python standard lib. This is to use both TOML reader and writer from the same source: PyPi.

kubernetes/src/charm.py
sinclert-canonical: That is not what the `charm-refresh` [docs](https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/log-format/) recommend.

machines/tests/integration/integration/test_backup_aws.py
sinclert-canonical: I was running into trouble with these snippets. It may have been related to a prior version of this PR code, but honestly, these scale-downs are unnecessary, and deviate from what we have in the K8s tests.

Better to prune them.

kubernetes/src/charm.py
astrojuanlu: Thanks for the pointer!

machines/src/mysql_vm_helpers.py
sinclert-canonical: Sorry, I thought you were referrencing the `tomli` package being introduced to the test suite (I got confused between refresh V3 PRs 3, 4 & 5). You are absolutely right: the `tomli` package is the one coming from PyPi, which is not listed as a main dependency. So for the production code, either we:

- Default to the Python 3.11+ `tomllib` standard library.
- Extend the dependencies with the `tomli` package.

astrojuanlu: Agreed with including `tomli` in the deps (as you just did in ca9d774) for future consistency with tomli_w

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -138,7 +139,7 @@ def __init__(
 LIBAPI = 0
 LIBPATCH = 103
 
-PYDEPS = ["mysql_shell_client ~= 0.7"]
+PYDEPS = ["charm_refresh ~= 3.1.1", "mysql_shell_client ~= 0.7"]
```
paulomach, line 142: Given we are not using this as charm libs, do we really need to set dependency here?

kubernetes/src/charm.py
```diff
@@ -94,13 +93,16 @@
...
+from log_rotation_setup import LogRotationSetup
 from mysql_k8s_helpers import MySQL, MySQLInitialiseMySQLDError
+from refresh import KubernetesMySQLRefresh
 from relations.mysql_provider import MySQLProvider
 from rotate_mysql_logs import RotateMySQLLogs, RotateMySQLLogsCharmEvents
-from upgrade import MySQLK8sUpgrade, get_mysql_k8s_dependencies_model
 from utils import compare_dictionaries, dotappend, generate_random_password, get_k8s_fqdn
 
 logger = logging.getLogger(__name__)
+logging.getLogger("httpx").setLevel(logging.WARNING)
```
paulomach, line 104: ~~on k8s helpers we set the same components log level to ERROR, maybe centralize here?~~ 
is removed!

kubernetes/tests/integration/helpers_ha.py
```diff
@@ -268,11 +268,13 @@ def scale_app_units(juju: Juju, app_name: str, num_units: int) -> None:
         ready=lambda status: len(status.apps[app_name].units) == num_units,
         timeout=20 * MINUTE_SECS,
     )
-    juju.wait(
-        ready=wait_for_apps_status(jubilant.all_active, app_name),
-        error=jubilant.any_blocked,
-        timeout=20 * MINUTE_SECS,
-    )
+
+    if num_units > 0:
```
taurus-forever, line 272: This is actually funny... what happens when:
 * I deploy charm rev A with MySQL 8.4.A
 * scale to 0
 * refresh to charm rev B (supposed to be shipped with MySQL 8.4.B)
 * scale to 3
New charm code will wake up with incompatible snap revision?

kubernetes/tests/spread/integration/test_async_replication_upgrade.py/task.yaml
```diff
@@ -2,6 +2,8 @@ summary: test_async_replication_upgrade.py
 environment:
   TEST_MODULE: high_availability/test_async_replication_upgrade.py
 execute: |
-  tox run -e integration -- "tests/integration/integration/$TEST_MODULE" --alluredir="$SPREAD_TASK/allure-results"
+  # TODO: Un-comment when we release a refresh V3 based charm to 8.4/edge
```
taurus-forever, line 5: Please create the next pulse Jira backlog ticket for this TODO. :-D

taurus-forever (commented):
LGTM, please review it with @paulomach as well!
Spotted intresting case in comments.

For the history: we can merge this without backward compatibility as there are no 8.4/stable releases yet. It cannot be ported to 8.0 AS IS.

taurus-forever (approved):

kubernetes/tests/integration/helpers_ha.py
sinclert-canonical: My intuition is that in Kubernetes, the old version of the **rock** will be used, unless the refresh procedure is accompanied with some resource updated logic. However, I think Machines code will download the correct **snap**, as such information is packed into the metadata that gets refreshed.

In any case, the code you are referencing is a test helper.

kubernetes/tests/spread/integration/test_async_replication_upgrade.py/task.yaml
sinclert-canonical: The follow-up [PR](https://github.com/canonical/mysql-operators/pull/149) uncomments the spread tasks (see [commit](https://github.com/canonical/mysql-operators/pull/149/changes/c929ffd69b761ba96c5127168d2ca0e1317deff4)). No need to create JIRA ticket.

sinclert-canonical:
Thanks @taurus-forever!

> For the history: we can merge this without backward compatibility as there are no 8.4/stable releases yet. It cannot be ported to 8.0 AS IS.

I do not think the plan is to backport refresh V3 to the `8.0/edge` branch, as it is, by design, an incompatible change. In principle, it will never be backported.

kubernetes/src/refresh.py
```diff
@@ -0,0 +1,108 @@
...
+
+        return all((
+            old_major == new_major,
+            old_minor == new_minor,
+        ))
+
+    @property
+    def app_units(self) -> set:
+        """The peer-related units in the application."""
+        return {self._charm.unit, *self._charm.peers.units}
```
paulomach:
```suggestion
        return self._charm.app_units
```

kubernetes/tests/spread/integration/test_async_replication_upgrade.py/task.yaml
```diff
@@ -2,6 +2,8 @@ summary: test_async_replication_upgrade.py
...
+  # tox run -e integration -- "tests/integration/integration/$TEST_MODULE" --alluredir="$SPREAD_TASK/allure-results"
+  exit 0
```
paulomach, line 7: 100% pass rate

machines/src/charm.py
```diff
@@ -111,11 +113,13 @@
     snap,
     snap_service_operation,
 )
+from refresh import MachinesMySQLRefresh
 from relations.mysql_provider import MySQLProvider
-from upgrade import MySQLVMUpgrade, get_mysql_dependencies_model
 from utils import compare_dictionaries, generate_random_password
 
 logger = logging.getLogger(__name__)
+logging.getLogger("httpx").setLevel(logging.WARNING)
```
paulomach, line 121: AFAIK this was to reduce noise on k8s only, due http{x,core} used by lightkube.

machines/src/refresh.py
```diff
@@ -0,0 +1,125 @@
...
+
+        return all((
+            old_major == new_major,
+            old_minor == new_minor,
+        ))
+
+    @property
+    def app_units(self) -> set:
+        """The peer-related units in the application."""
+        return {self._charm.unit, *self._charm.peers.units}
```
paulomach: same as for k8s

machines/tests/integration/integration/test_backup_aws.py
paulomach: this is from an era where default GH runner was smaller

paulomach (approved):
Nothing to say about function nor form. Just pointers on maybe unnecessary code.
Great work

kubernetes/lib/charms/mysql/v0/mysql.py
sinclert-canonical: You are right here: 100% unnecessary. Let's leave it as is until we cleanup all the charm-lib stuff.

kubernetes/src/refresh.py
sinclert-canonical: Good catch. Simplified in commit https://github.com/canonical/mysql-operators/pull/148/commits/1ac2d17ab1078c262f7d1a67dbaa3b88059ab5bc.

kubernetes/tests/spread/integration/test_async_replication_upgrade.py/task.yaml
sinclert-canonical: Yes, this is necessary until a refresh V3 charm is published to the `8.4/edge` branch. However, I already uncommented + tested it in the follow-up [PR](https://github.com/canonical/mysql-operators/pull/149).

machines/src/charm.py
sinclert-canonical: Exactly. This is the log level recommended by charm-refresh [docs](https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/httpx-logger-level/).

machines/src/refresh.py
sinclert-canonical: Good catch. Simplified in commit https://github.com/canonical/mysql-operators/pull/148/commits/1ac2d17ab1078c262f7d1a67dbaa3b88059ab5bc.

sinclert-canonical:
> The hard-coded charm key within the refresh versions YAML files needs to be reverted before merging.

Temporal changes reverted in commit https://github.com/canonical/mysql-operators/pull/148/commits/231227f083af77f6ea73b25572d2281a4203ed37.

Issue #93, opened 2026-02-17
Removing a second application breaks mysql-k8s primary / secondary endpoints


gboutry:
## Steps to reproduce

```
juju add-model squirrel ck8s --config logging-config="<root>=INFO;unit=DEBUG"
juju deploy mysql-k8s mysql --trust --channel 8.0/stable --revision 343
juju deploy mysql-test-app req-a --config database_name=req-a
juju deploy mysql-test-app req-b --config database_name=req-b
juju integrate mysql:database req-a:database
juju integrate mysql:database req-b:database
juju wait-for unit mysql/0 --query='(workload-status=="active" && agent-status=="idle")'
juju wait-for unit req-a/0 --query='(workload-status=="active" && agent-status=="idle")'
juju wait-for unit req-b/0 --query='(workload-status=="active" && agent-status=="idle")'
juju remove-application req-b --destroy-storage --no-prompt 
```

## Expected behavior
Mysql should continue to work even if a requiring application is gone

## Actual behavior
Mysql deletes it's k8s endpoints, breaking name resolution

## Versions

Operating system: Ubuntu

Juju CLI: 3.6.14

Juju agent: 3.6.14

Charm revision: 343

k8s:   v1.32.11  4754  1.32-classic/stable  canonical✓  classic

## Log output

Juju debug log: 

```
2026-02-17T10:47:54.772Z [container-agent] 2026-02-17 10:47:54 INFO juju.worker.uniter.operation runhook.go:186 ran "database-relation-changed" hook (via hook dispatching script: dispatch)                                                             
2026-02-17T10:47:55.298Z [container-agent] 2026-02-17 10:47:55 DEBUG juju-log database-peers:0: ops 2.15.0 up and running.                                                                                                                               
2026-02-17T10:47:55.515Z [container-agent] 2026-02-17 10:47:55 DEBUG juju-log database-peers:0: no relation on 'tracing': tracing not ready                                                                                                              
2026-02-17T10:47:55.521Z [container-agent] 2026-02-17 10:47:55 DEBUG juju-log database-peers:0: Starting root trace with id='38131d575f5e444ddd364264e1dbc1c1'.                                                                                          
2026-02-17T10:47:55.525Z [container-agent] 2026-02-17 10:47:55 DEBUG juju-log database-peers:0: Emitting Juju event database_peers_relation_changed.                                                                                                     
[...]

PR #152, opened 2026-03-12
[DPE-9469] 8.0 - Fix relation-broken services

sinclert-canonical:
This PR addresses issue https://github.com/canonical/mysql-operators/issues/93 by removing the K8s services only when the number of apps under the same relation is 0.

### DIFF
--- a/kubernetes/src/relations/mysql_provider.py
+++ b/kubernetes/src/relations/mysql_provider.py
@@ -267,8 +267,7 @@ class MySQLProvider(Object):
             # https://bugs.launchpad.net/juju/+bug/1979811
             return
 
-        if len(self.model.relations[DB_RELATION_NAME]) == 1:
-            # remove kubernetes service when last relation is removed
+        if len(self.model.relations[DB_RELATION_NAME]) == 0:
             self.charm.k8s_helpers.delete_endpoint_services(["primary", "replicas"])
 
         relation_id = event.relation.id
--- a/kubernetes/tests/integration/integration/relations/test_database.py
+++ b/kubernetes/tests/integration/integration/relations/test_database.py
@@ -119,3 +119,57 @@ def test_relation_broken(juju: Juju):
         timeout=15 * MINUTE_SECS,
         delay=2,
     )
+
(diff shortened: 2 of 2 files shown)


kubernetes/src/relations/mysql_provider.py
```diff
@@ -267,8 +267,7 @@ def _on_database_broken(self, event: RelationBrokenEvent) -> None:
             # https://bugs.launchpad.net/juju/+bug/1979811
             return
 
-        if len(self.model.relations[DB_RELATION_NAME]) == 1:
-            # remove kubernetes service when last relation is removed
+        if len(self.model.relations[DB_RELATION_NAME]) == 0:
```
paulomach, line 270: My doubt here is if there's some change in behavior in juju/ops - IIRC, when the event was triggered, i.e. on the relation-broken event, the relation being remove was still available through `self.model.relations[DB_RELATION_NAME]`, hence the `1` and not `0`

paulomach: Anyway, this is not a blocker, as having the service enable even without active relation is harmless

paulomach (approved):

kubernetes/src/relations/mysql_provider.py
sinclert-canonical: We never tested it. We have a few tests actively involving the MySQL Test app, but either:

- We test having multiple units of the same app (see [test](https://github.com/canonical/mysql-operators/blob/df35a7d43733b0d158886571f0156abdc6d40103/kubernetes/tests/integration/integration/relations/test_database.py#L103-L121)).
- We test having multiple apps, but never removing their relations (see [tests](https://github.com/canonical/mysql-operators/blob/df35a7d43733b0d158886571f0156abdc6d40103/kubernetes/tests/integration/integration/test_multi_relations.py)).

---

The only possible behavior change that comes to mind is whether this is different between Juju 2.9 and 3.6, but we should be able to see that when the test suite runs, as the remaining MySQL Test app would lose all connectivity, making the `start-continuous-writes` action to fail.

astrojuanlu (approved):

PR #153, opened 2026-03-12
[DPE-9469] 8.4 - Fix relation-broken services

sinclert-canonical:
This PR addresses issue https://github.com/canonical/mysql-operators/issues/93 by removing the K8s services only when the number of apps under the same relation is 0.

### DIFF
--- a/kubernetes/src/relations/mysql_provider.py
+++ b/kubernetes/src/relations/mysql_provider.py
@@ -264,8 +264,7 @@ class MySQLProvider(Object):
             # https://bugs.launchpad.net/juju/+bug/1979811
             return
 
-        if len(self.model.relations[DB_RELATION_NAME]) == 1:
-            # remove kubernetes service when last relation is removed
+        if len(self.model.relations[DB_RELATION_NAME]) == 0:
             self.charm.k8s_helpers.delete_endpoint_services(["primary", "replicas"])
 
         relation_id = event.relation.id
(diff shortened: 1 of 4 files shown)


paulomach (approved):

astrojuanlu (approved):

kubernetes/src/relations/mysql_provider.py
```diff
@@ -264,8 +264,7 @@ def _on_database_broken(self, event: RelationBrokenEvent) -> None:
             # https://bugs.launchpad.net/juju/+bug/1979811
             return
 
-        if len(self.model.relations[DB_RELATION_NAME]) == 1:
-            # remove kubernetes service when last relation is removed
+        if len(self.model.relations[DB_RELATION_NAME]) == 0:
```
taurus-forever, line 267: Why do we remove K8s serviced on relation removal IF they should stay untill DB app/pods exist. So delete it only on DB app removal?

sinclert-canonical: I think is because of the chosen strategy to deal with the service life-cycle:

- They get **created** when a client application relates.
- They get **removed** when no client application is related.

Now, why such strategy, instead of creating the K8s services from the very beginning and bind them to the life-time of the MySQL application? I do not know. Maybe something about the Juju TearDown event? 🤷🏻

sinclert-canonical: See issue https://github.com/canonical/mysql-operators/issues/157.

PR #156, opened 2026-03-13
8.4 - Fix: Do not decode tls-ca-chain received via S3 relation

### LINKED ISSUE #92: [k8s] [vm] mysql charms do not work with S3 clouds with custom CA

theoctober19th:
## Issue
Part 1 of the solution to fix https://github.com/canonical/mysql-operators/issues/92.

## Solution
Do not decode tls-ca-chain received via S3 relation since it is already decoded.

## Checklist

### DIFF
--- a/kubernetes/lib/charms/mysql/v0/s3_helpers.py
+++ b/kubernetes/lib/charms/mysql/v0/s3_helpers.py
@@ -88,7 +88,7 @@ def _get_bucket(s3_parameters: dict) -> boto3.resources.base.ServiceResource:
 
     with tempfile.NamedTemporaryFile() if ca_chain else nullcontext() as ca_file:
         if ca_file:
-            ca = "\n".join([base64.b64decode(s).decode() for s in ca_chain])
+            ca = "\n".join(ca_chain)
             ca_file.write(ca.encode())
             ca_file.flush()
 
--- a/machines/lib/charms/mysql/v0/s3_helpers.py
+++ b/machines/lib/charms/mysql/v0/s3_helpers.py
@@ -14,7 +14,6 @@
 
 """S3 helper functions for the MySQL charms."""
 
-import base64
 import logging
 import pathlib
 import tempfile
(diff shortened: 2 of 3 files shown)


sinclert-canonical (approved):
Thanks Bikalpa! A couple of notes for reviewers:

- This fix amends community contributed PR https://github.com/canonical/mysql-operator/pull/396.
- This fix needs to be merged alongside PR https://github.com/canonical/mysql-operators/pull/105 in order to fix the usage of backups alongside TLS.

paulomach (approved):

astrojuanlu (approved):
Thanks @theoctober19th !

Issue #159, opened 2026-03-17
Review and update guidance for upgrades with server + router


astrojuanlu:
While investigating #151, I was trying to follow our official documentation https://canonical-charmed-mysql.readthedocs-hosted.com/how-to/refresh/single-cluster/refresh-single-cluster/ and found some faulty situations.

Upgrading the server from rev 255 to rev 343 (8.0/stable):

| Server units | Router units | Result                                                                             |
|--------------|--------------|------------------------------------------------------------------------------------|
| 1            | 0            | Upgrade succeeds                                                                   |
| 2            | 0            | Upgrade succeeds                                                                   |
| 1            | 1            | Upgrade succeeds but ⚠️ "update-status" check fails in router, unit ends up in error |
| 2            | 1            | Upgrade gets stuck 💥                                                              |
| 3            | 1            | Upgrade succeeds                                                                   |
| 4            | 1            | Upgrade succeeds                                                                   |

The (2, 1) scenario from rev 240 to rev 255 and it also got stuck. So it's not something new of MySQL 8.0.44.

The docs say 

> **Step 2: Scale up (optional)**
>
> With Charmed MySQL for Kubernetes, it is recommended to scale the application up by one unit before starting the upgrade process.

but somewhere else https://canonical-charmed-mysql.readthedocs-hosted.com/how-to/scale/ they also say

> It is recommended to use an odd number to prevent a split-brain scenario.

astrojuanlu:
@sinclert-canonical proposes:

> ### Step 2: Scale up
>
> With Charmed MySQL for Kubernetes, it is mandatory to have a minimum of 3 healthy units before starting the upgrade process

astrojuanlu:
Fixed in #160.

PR #160, opened 2026-03-17
[MISC] Update refresh single-cluster steps

sinclert-canonical:
This PR addresses issue https://github.com/canonical/mysql-operators/issues/159, where we discovered that it is absolutely mandatory to have at least 3 MySQL Server units when refreshing, if a MySQL Router application is connected. In order to make the documentation generic enough, we state that a minimum of 3 units is required.

I also updated the commands, to show-case how to scale an application both in VM & K8s case.

### DIFF
--- a/docs/how-to/refresh/single-cluster/refresh-single-cluster.md
+++ b/docs/how-to/refresh/single-cluster/refresh-single-cluster.md
@@ -55,7 +55,7 @@ For this example, the current revision is `182`. Store it safely to use in case 
 
 ## Step 2: Scale up
 
-With Charmed MySQL, it is mandatory to have at least 3 units. This ensures that the MySQL cluster can refresh regardless of whether a MySQL Router application is connected, as the quorum criteria within the cluster changes on this condition.
+It is mandatory to have at least 3 units of MySQL before refreshing. This ensures that the MySQL cluster can refresh regardless of whether a MySQL Router application is connected, as the quorum criteria within the cluster varies based on this condition.
 
 In case of failure, having extra units will ease a future rollback procedure without disrupting service. To scale up the application:
 
--- a/docs/how-to/refresh/single-cluster/refresh-single-cluster.md
+++ b/docs/how-to/refresh/single-cluster/refresh-single-cluster.md
@@ -24,7 +24,7 @@ To upgrade Juju, see {ref}`upgrade-juju`
 ## Step 1: Record revision information
 
 ```{note}
-This step is only valid when deploying from [Charmhub](https://charmhub.io/mysql). 
+This step is only valid when deploying from [Charmhub](https://charmhub.io/mysql).
 
 If a [local charm](https://juju.is/docs/sdk/deploy-a-charm) is deployed (revision is small, e.g. 0-10), make sure the proper/current local revision of the `.charm` file is available BEFORE going further. You might need it for a rollback.
 ```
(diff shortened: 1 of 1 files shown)


astrojuanlu (approved):
👍🏼

docs/how-to/refresh/single-cluster/refresh-single-cluster.md
```diff
@@ -53,19 +53,35 @@ Machine  State    Address         Inst id         Series  AZ  Message
 
 For this example, the current revision is `182`. Store it safely to use in case of rollback!
 
-## Step 2: Scale up (optional)
+## Step 2: Scale up
 
-With Charmed MySQL for Kubernetes, it is recommended to scale the application up by one unit before starting the upgrade process.
+With Charmed MySQL, it is mandatory to have at least 3 units. This ensures that the MySQL cluster can refresh regardless of whether a MySQL Router application is connected, as the quorum criteria within the cluster changes on this condition.
```
taurus-forever: re: `With Charmed MySQL, it is mandatory to have at least 3 units.`
As a newbie, I read it as 1 and 2 units are not allowed.
I believe 1 unit deployment is still allowed. Nope?


Maybe `Charmed MySQL supports single unit deployment and clustered one with 3-9 units in cluster. Two units deployment is not supported for production usage but available for experiments.`

sinclert-canonical: > As a newbie, I read it as 1 and 2 units are not allowed. I believe 1 unit deployment is still allowed. Nope?

This is the matrix that we care about:

| MySQL Server units | MySQL Router connected? | Result                                                   |
| -----------------------------| -------------------------------------- |----------------------------------------------------|
| 1                                 | No                                            | ✅ (MySQL Server cannot rollback) |
| 1                                 | Yes                                           | 💥 (MySQL Router raises an error)  |
| 2                                 | No                                            | ✅                                                          |
| 2                                 | Yes                                           | 💥 (MySQL Server raises an error)  |
| 3                                 | No                                            | ✅                                                          |
| 3                                 | Yes                                           | ✅                                                          |

As you can see, the results greatly varies depending on having a MySQL Router application related. This happens because MySQL Router introduces some HA requirements that a MySQL Server cluster with less than 3 units does not reach.

Given that we are explicitly recommending to have a MySQL Router app deployed next to the MySQL Server, in order to minimize disconnections, when doing a refresh (see [docs](https://github.com/canonical/mysql-operators/blob/08a456e09f5b117ac1aa09ef07382066e47c55be/docs/how-to/refresh/single-cluster/refresh-single-cluster.md?plain=1#L22)), I think requiring a minimum of 3 MySQL Server units is the sanest approach.

---

**Disclaimer:**

All testing was done with MySQL Server K8s. I assume MySQL Server HA requirements will be the the same in the VM substrate.

a-velasco: I agree with @sinclert-canonical's approach. We can consider adding more nuanced information about quorum criteria with/without MySQL Router in another part of the documentation (I already have some places in mind) - but this seems like a good compromise between a safe recommendation and enough context.

Minor suggestion - I think the phrasing would be a bit more clear like this.

```suggestion
It is mandatory to have at least 3 units of MySQL before refreshing. This ensures that the MySQL cluster can refresh regardless of whether a MySQL Router application is connected, as the quorum criteria within the cluster varies based on this condition.
```

a-velasco (approved):

PR #173, opened 2026-03-23
[DPE-8295] Refresh V3 (III.2): Fix automatic tagging

sinclert-canonical:
This PR is a hot-fix over the previous [PR](https://github.com/canonical/mysql-operators/pull/148) (the one introducing the whole refresh V3 functionality), so that the automatic tagging is triggered before the building process starts.

### Disclaimer

Depending on `charmcraftlocal` goes against the general guideline of using charmcraft wrappers, but there is not way to introduce refresh V3 vanilla implementation into a mono-repo setup otherwise. We **need** to do the tagging before actually starting the packing process, as vanilla charmcraft does not have access to the `.git` folder in the parent directory.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,13 +19,13 @@ this operator.
   the `8.4/edge` branch. This also avoids merge commits and creates a linear Git commit history.
 
 ## Develop
-Install `yq`, `tox`, `poetry`, and `charmcraftcache`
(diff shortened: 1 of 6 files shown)


astrojuanlu (approved):
👍🏼

paulomach (approved):
Thank you.

@carlcsaposs-canonical , we need to think what can be done on `ccl` to identify a monorepo without a local package.

PR #213, opened 2026-04-06
[MISC] Various CI fixes

astrojuanlu:
## Issue
- Fix publishing of Allure results
- Fix recurring "relation data-integrator1:mysql mysql:database is dying, but not yet removed (already exists)"

## Checklist

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -148,7 +148,7 @@ jobs:
         # Only upload results from one spread system & one spread variant
         # Allure can only process one result per pytest test ID. If parameterization is done via
         # spread instead of pytest, there will be overlapping pytest test IDs.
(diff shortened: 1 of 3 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -148,7 +148,7 @@ jobs:
         # Only upload results from one spread system & one spread variant
         # Allure can only process one result per pytest test ID. If parameterization is done via
         # spread instead of pytest, there will be overlapping pytest test IDs.
-        if: ${{ (success() || (failure() && steps.spread.outcome == 'failure')) && startsWith(matrix.job.spread_job, 'github-ci:ubuntu-24.04:') && github.event_name == 'schedule' && github.run_attempt == '1' }}
```
astrojuanlu, line 151: Carl confirmed that we'd need to keep the selector here, otherwise we'd have overlapping test case IDs and Allure would break

sinclert-canonical: > Carl confirmed that we'd need to keep the selector here

Too bad, I would rather have the Ubuntu runner information out of the selector, so that we can change runners in a central place, without needing to tweak the workflows...

sinclert-canonical (approved):

paulomach (approved):

PR #233, opened 2026-04-10
[DPE-9791] Handle stale primary on crash when joining peer

### LINKED ISSUE #227: [k8s/vm] Self heal on primary crash while other unit joining the cluster
## Steps to reproduce

On sunbeam deployment, while a unit was joining the cluster, if the primary crashes, the primary fails to recover and new units cannot rejoin the cluster.

1. TBD

## Expected behavior

Primary recover and units resume joining the cluster

### Primary recover

Primary cannot recover due to [validation to check if any other unit is already](https://github.com/canonical/mysql-operators/blob/921a790fc4fc43d6db3a6d30ec9b29e664d1c5d6/kubernetes/src/charm.py#L846) in the cluster fails, as the joining unit already reports itself as part of the cluster:

Image

After manual call to `dba.reboot_from_complete_outage`, the cluster recover only with the primary, so the joining unit is not yet accounted for:

Image

### Resume joining units

The joining unit(s) will be blocked from rejoining the instance, due the locking table (used to coordinate instance additon/removal) still holding the lock for the joining instance in the primary:

Image

Manual resetting the lock table allow procedure to resume.

## Versions

Operating system: ubuntu noble

Juju CLI: juju-3.6.21

Juju agent: juju-3.6.21

Charm revision: <=400

Microk8s: *
LXD: *

## Additional context

Implementation suggestion:
1. Review reboot_from_complete outage conditions frame this case
2. On reboot_from_complete treatment, reset unit_operations table

paulomach:
Fix #227
Depends on https://github.com/canonical/mysql-shell-client/pull/54 - ✅


kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2161,6 +2161,18 @@ def _release_lock(self, executor: BaseExecutor, unit_label: str, unit_task: str)
         else:
             return True
 
+    def release_all_locks(self, executor: BaseExecutor):
```
sinclert-canonical: I would suggest making the method private for now, same as `_acquire_lock` and `_release_lock` ones, given that it is only used within the lib itself.

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2372,6 +2384,8 @@ def reboot_from_complete_outage(self) -> None:
         """Wrapper for reboot_cluster_from_complete_outage command."""
         try:
             self._cluster_client_tcp.reboot_cluster(self.cluster_name)
+            executor = self._build_cluster_tcp_executor(self.instance_address)
+            self.release_all_locks(executor)
```
sinclert-canonical: I would suggest building the executor outside of the `try-except` block, similar to how other methods in this class do it. It would allow us to focu our attentiont into the lines that _really_ can yield exceptions.

kubernetes/src/charm.py
```diff
@@ -881,21 +877,25 @@ def _handle_potential_cluster_crash_scenario(self) -> bool:  # noqa: C901
                 self.peers.data[unit].get("member-state", "unknown") for unit in self.peers.units
             }
 
+            peers_waiting = all_states == {"waiting"}
             # Add state 'offline' for this unit (self.peers.unit does not include this unit)
-            if (all_states | {"offline"} == {"offline"} and self.unit.is_leader()) or (
-                only_single_uninitialized_node_across_cluster and all_states == {"waiting"}
-            ):
-                # All instance are off, reboot cluster from outage from the leader unit
+            all_offline = all_states | {"offline"} == {"offline"}
```
sinclert-canonical, line 882: I love this refactor ❤️

sinclert-canonical (commented):
LGTM.

Please consider that every comment left in the `mysql` charmlib should be addressed both in K8s and VM versions

kubernetes/lib/charms/mysql/v0/mysql.py
paulomach: 936e687c

kubernetes/lib/charms/mysql/v0/mysql.py
paulomach: 936e687c

kubernetes/src/charm.py
```diff
@@ -881,21 +877,25 @@ def _handle_potential_cluster_crash_scenario(self) -> bool:  # noqa: C901
...
 
+            if (all_offline and self.unit.is_leader()) or (peers_waiting):
```
astrojuanlu: nit:


```suggestion
            if (all_offline and self.unit.is_leader()) or peers_waiting:
```

astrojuanlu (approved):

kubernetes/src/charm.py
paulomach: b63302c5

sinclert-canonical (approved):

PR #238, opened 2026-04-14
[MISC] Reintroduce `--model` parameter for pytest runs

astrojuanlu:
## Issue

I used to find the `--model` parameter very useful when running tests, so I'd like to propose reintroducing it. Example use cases:
- Launch the tests to a new model (`testing2`, or just an auto-generated one) while you destroy and recreate `testing` (which in the case of K8s takes quite a bit of time, or may be even blocked by things like https://github.com/juju/juju/issues/22075)
- Launch the tests in 2 different models to compare logs

I've been keeping dirtier versions of this to myself, but it's annoying to have those uncommitted changes all the time, so I'd like to include this in the source tree. This is much simpler than what we had before #62, in that the parameter is completely optional and has a default value of `testing`.

Happy to backport this to 8.0/edge if approved.

## Checklist

### DIFF
--- a/kubernetes/tests/integration/conftest.py
+++ b/kubernetes/tests/integration/conftest.py
@@ -25,6 +25,17 @@ def pytest_configure(config):
     config.option.log_date_format = "%b %d %H:%M:%S"
 
 
+def pytest_addoption(parser):
+    """Adds command line parameter ``--model`` (see help for details)."""
+    parser.addoption(
+        "--model",
+        action="store",
+        default="testing",
+        help="model name or ':auto:' for temporary model, default to 'testing'",
+        required=False,
+    )
+
+
 @pytest.fixture(scope="session")
 def charm():
     # Return str instead of pathlib.Path since python-libjuju's model.deploy(), juju deploy, and
@@ -64,5 +75,12 @@ def cloud_configs_gcp() -> tuple[dict[str, str], dict[str, str]]:
 
 
 @pytest.fixture(scope="module")
-def juju() -> jubilant.Juju:
-    return jubilant.Juju(model="testing")
+def juju(request: pytest.FixtureRequest):
+    """Pytest fixture that yields a new :meth:`jubilant.Juju` object."""
+    model = request.config.getoption("--model")
+
+    if model == ":auto:":
+        with jubilant.temp_model() as juju:
+            yield juju
+    else:
+        yield jubilant.Juju(model=model)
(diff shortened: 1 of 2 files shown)


paulomach:
> ## Issue
> 
> I used to find the `--model` parameter very useful when running tests, so I'd like to propose reintroducing it. Example use cases:
> 
>     * Launch the tests to a new model (`testing2`, or just an auto-generated one) while you destroy and recreate `testing` (which in the case of K8s takes quite a bit of time, or may be even blocked by things like [Juju stuck at destroying k8s model juju/juju#22075](https://github.com/juju/juju/issues/22075))
> 
>     * Launch the tests in 2 different models to compare logs
> 
> 
> I've been keeping dirtier versions of this to myself, but it's annoying to have those uncommitted changes all the time, so I'd like to include this in the source tree. This is much simpler than what we had before #62, in that the parameter is completely optional and has a default value of `testing`.
> 
> Happy to backport this to 8.0/edge if approved.
> ## Solution
> ## Checklist
> 
>     * [ ]  I have added or updated any relevant documentation.
> 
>     * [ ]  I have cleaned any remaining cloud resources from my accounts.

How about do the full thing and use [pytest-jubilant](https://pypi.org/project/pytest-jubilant/) instead. It has extra nice fixtures we don't need to maintain, including one to accept juju model name

astrojuanlu:
I'm not opposed to it, but I'm not using any of those features and I believe @sinclert-canonical was in favour of reducing the number of dependencies. As long as we don't have to touch all our `task.yaml` files again, I'm fine either way.

paulomach:
> I'm not opposed to it, but I'm not using any of those features and I believe @sinclert-canonical was in favour of reducing the number of dependencies. As long as we don't have to touch all our `task.yaml` files again, I'm fine either way.

@sinclert-canonical ?

sinclert-canonical:
> @sinclert-canonical ?

I am in favor of reducing the number of dependencies.

AFAIK, the Kafka team, who also migrated most of their integration tests to Jubilant, are manually defining a Pytest option to expose the `--model` argument (presented during the last Tech Talk). The proposed approach LGTM.

sinclert-canonical (approved):
Is `:auto:` a common way to deal with temporal models? I have not seen that before.

astrojuanlu:
> Is `:auto:` a common way to deal with temporal models? I have not seen that before.

No, I made it up 😄

paulomach (approved):

PR #235, opened 2026-04-13
[DPE-9805] Fix rollback incompatible tests

astrojuanlu:
## Issue

- A `check=True` was missing when initialising the mysqld config, otherwise `CalledProcessError` would never be raised
- Simplified the rollback test to deploy old charm instead of old snap

Fix #234

## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.

PR #182, opened 2026-03-26
[DPE-7908] Initialise MySQL data directory in charm (backport)

astrojuanlu:
## Issue

Backports https://github.com/canonical/mysql-operators/pull/177

## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.

PR #246, opened 2026-04-16
[MISC] Remove feature rollback from incompatible

paulomach:
This feature is used for the edge case when a rollback is needed
if a refresh from mysql version <=8.0.34 fails (usually it does not)
and there is more than one unit available to recover from.

The feature adds a lot of complexity and test is flaky.
Recent PR #182 broke the feature, as it we would need to initialise or reinstall the snap
to create the initial data for mysqld, depending on the snap revision.
See PR #235 to what that could potentially look like.

Additionally, given mysql-8.0 is EOL, we are sure there will not be further
minor versions bumps that will break data dir compatibility (as was the case with
the 8.0.34 -> 8.0.35).


astrojuanlu (approved):
Reasonable.

Should we add a short note to https://canonical-charmed-mysql.readthedocs-hosted.com/8.0/how-to/refresh/single-cluster/roll-back-single-cluster/ ?

paulomach:
> Reasonable.
> 
> Should we add a short note to https://canonical-charmed-mysql.readthedocs-hosted.com/8.0/how-to/refresh/single-cluster/roll-back-single-cluster/ ?

Added that and removed the feature in k8s

sinclert-canonical (approved):
🔥 LGTM. I have an uncomfortable question to ask though:

In the PR description you mentioned:

> _PR https://github.com/canonical/mysql-operators/pull/182 broke the feature, as it we would need to initialise or reinstall the snap to create the initial data for mysqld, depending on the snap revision_.

Is that not the case for every single rollback? How can we guarantee that the previous charm code version would have the _initialize_ logic already?

paulomach:
> 🔥 LGTM. I have an uncomfortable question to ask though:
> 
> In the PR description you mentioned:
> 
> > _PR #182 broke the feature, as it we would need to initialise or reinstall the snap to create the initial data for mysqld, depending on the snap revision_.
> 
> Is that not the case for every single rollback? How can we guarantee that the previous charm code version would have the _initialize_ logic already?

No, it was particular to the feature, as the initialization is only done due the data directory being wiped out and reinitialized with the prior mysql version, ensuring the compatibility.

In case of a normal rollback, data is kept.

paulomach:
Remaining tests fails are unrelated. Merging as is

PR #248, opened 2026-04-16
[DPE-7899] 8.4 - Migrate to TLS v4 (I)

sinclert-canonical:
This PR migrates the handling of TLS certificates to use version 4 of the `tls-certificates` interface.

The library used to deal with version 4 of such interface was initially published as a _charmlib_, but then ported to a pure Python package within the charmlibs repository, to be published and consumed via PyPi. Despite the package current version (`1.X.Y`), it is shipping version 4 of the interface.

### Approach:

The new interface version allow us to simplify the current logic by skipping any _certificate expiration_ event handling. This is now managed by the provider automatically, so the only event we need to handle is `CertificateAvailable`.

### Future work:

1. Enable _peer-to-peer_ encryption by enabling the group replication SSL.
2. Define the necessary logic to parse a manually configured private key for TLS certificates generation.


kubernetes/src/charm.py
```diff
@@ -356,22 +356,24 @@ def get_unit_address(self, unit: Unit, relation_name: str = PEER) -> str:
...
         Translate juju unit name to resolvable hostname.
         """
-        unit_hostname = self.get_unit_hostname(unit.name)
-        unit_dns_domain = get_k8s_fqdn(self.get_unit_hostname(unit.name))
+        try:
+            unit_hostname = self.get_unit_hostname(unit.name)
+            unit_dns_domain = get_k8s_fqdn(self.get_unit_hostname(unit.name))
+        except RuntimeError:
+            logger.warning("Unit DNS domain name is not propagated yet")
+            return ""
```
sinclert-canonical, line 367: This is required so that the TLS class initialization does not crash when the charm does not has any of the expected relations (_database_, _database-peers_, _replication_, _replication-offer_). This usually happens at the very early stages of a charm deployment, or during unit tests.

machines/src/charm.py
```diff
@@ -812,6 +812,9 @@ def workload_initialise(self) -> None:
 
     def get_unit_address(self, unit: Unit, relation_name: str) -> str:
         """Get the IP address of a specific unit."""
+        if not self.peers:
+            return ""
```
sinclert-canonical, line 818: This is required so that the TLS class initialization does not crash when the charm does not has any of the expected relations (_database_, _database-peers_, _replication_, _replication-offer_). This usually happens at the very early stages of a charm deployment, or during unit tests.

kubernetes/src/relations/tls.py
```diff
@@ -0,0 +1,204 @@
...
+class TLS(Object):
+    """In this class we manage certificates relation."""
+
+    refresh_tls_certificates_event = EventSource(RefreshTLSCertificatesEvent)
+
+    def __init__(self, charm: "MySQLOperatorCharm"):
+        super().__init__(charm, "certificates")
+        self.charm = charm
+        self.unit_name = charm.unit.name.replace("/", "-")
+        self.unit_hostname = f"{self.unit_name}"
```
paulomach:
```suggestion
        self.unit_hostname = self.unit_label
```

kubernetes/src/charm.py
paulomach: Should we defer instead. I not sure how cert request behave with a `""` SAN

kubernetes/src/relations/tls.py
```diff
@@ -0,0 +1,204 @@
...
+        try:
+            self.charm._mysql.setup_client_tls(
+                ca_path=f"{MYSQL_DATA_DIR}/{TLS_SSL_CA_FILE}",
+                key_path=f"{MYSQL_DATA_DIR}/{TLS_SSL_KEY_FILE}",
+                cert_path=f"{MYSQL_DATA_DIR}/{TLS_SSL_CERT_FILE}",
+                require_tls=True,
+            )
+            self.charm._mysql.kill_client_sessions()
+        except MySQLTLSSetupError:
+            logger.error("Failed to enable TLS configuration.")
```
paulomach: This will silently fail, we need to raise (to use juju retry) or block (so user re-de relation)

kubernetes/src/charm.py
```diff
@@ -356,22 +356,24 @@ def get_unit_address(self, unit: Unit, relation_name: str = PEER) -> str:
 
         Translate juju unit name to resolvable hostname.
         """
-        unit_hostname = self.get_unit_hostname(unit.name)
-        unit_dns_domain = get_k8s_fqdn(self.get_unit_hostname(unit.name))
+        try:
+            unit_hostname = self.get_unit_hostname(unit.name)
+            unit_dns_domain = get_k8s_fqdn(self.get_unit_hostname(unit.name))
```
paulomach:
```suggestion
            unit_hostname = self.get_unit_hostname(unit.name)
            unit_dns_domain = get_k8s_fqdn(unit_hostname)
```

paulomach (commented):
Nice. Couple open questions, minor catches

kubernetes/src/relations/tls.py
sinclert-canonical: Done in commit https://github.com/canonical/mysql-operators/pull/248/commits/a0ee3ac883bed61a69337714d2020792663a48f1.

kubernetes/src/relations/tls.py
sinclert-canonical: Good point. I am restoring the previous way of dealing with this exception: to set a `Blocked` status.

Done in commit https://github.com/canonical/mysql-operators/pull/248/commits/a0ee3ac883bed61a69337714d2020792663a48f1.

kubernetes/src/charm.py
sinclert-canonical: Done in commit https://github.com/canonical/mysql-operators/commit/a0ee3ac883bed61a69337714d2020792663a48f1.

kubernetes/src/charm.py
sinclert-canonical: I am not entirely sure how to do so: the function does not received any event to defer upon 🤔 

Regardless, these host-names will only be used when the group-replication encryption mode is set to `VERIFY_IDENTITY` (see [docs](https://dev.mysql.com/doc/refman/8.4/en/group-replication-secure-socket-layer-support-ssl.html)), which we are far from using. The code is here to mimic PostgreSQL way of dealing with certificate generation.

sinclert-canonical:
🗞️  PR branch rebased from `8.4/edge`. This is ready for review again.

kubernetes/src/charm.py
paulomach: Good point. I'd say then to not include the empty str in the SAN set.

paulomach (approved):
No blocker from my side! Nice work

kubernetes/src/charm.py
sinclert-canonical: Avoiding the empty string is already managed [here](https://github.com/canonical/mysql-operators/blob/5758b1563a4719ef2eaed8ce45c46db0029d8c54/kubernetes/src/relations/tls.py#L94), similarly to PostgreSQL implementation (see [code](https://github.com/canonical/postgresql-k8s-operator/blob/9e23ab85ae3200d8ce1f836716a8c8c362b2ad9f/src/relations/tls.py#L58)).

machines/src/charm.py
astrojuanlu: In fact, we have several situations like these in our handlers, type checkers are especially noise about those

kubernetes/src/charm.py
astrojuanlu: What about moving this try...except one level up then? Not a fan of that `return ""`

machines/src/relations/tls.py
```diff
@@ -0,0 +1,175 @@
...
+
+        return client_addresses
+
+    def _get_client_tls_files(self) -> tuple[str | None, str | None, str | None]:
+        """Prepare TLS files in special MySQL way.
+
+        MySQL needs three files:
+        — CA file should have a full chain.
+        — Key file should have private key.
+        — Certificate file should have certificate without certificate chain.
```
astrojuanlu:
```suggestion
        - CA file should have a full chain.
        - Key file should have private key.
        - Certificate file should have certificate without certificate chain.
```

?

machines/src/relations/tls.py
```diff
@@ -0,0 +1,175 @@
...
+        try:
+            self.charm._mysql.setup_client_tls(
+                ca_path=f"{MYSQL_DATA_DIR}/client_{TLS_SSL_CA_FILE}",
+                key_path=f"{MYSQL_DATA_DIR}/client_{TLS_SSL_KEY_FILE}",
+                cert_path=f"{MYSQL_DATA_DIR}/client_{TLS_SSL_CERT_FILE}",
+                require_tls=True,
+            )
+            self.charm._mysql.kill_client_sessions()
+        except MySQLTLSSetupError:
+            logger.error("Failed to enable TLS configuration.")
```
astrojuanlu: Any problem with retaining the original context for debugging?


```suggestion
        except MySQLTLSSetupError as exc:
            logger.error(f"Failed to enable TLS configuration: {exc}")
```

machines/src/relations/tls.py
```diff
@@ -0,0 +1,175 @@
...
+            logger.debug("Unit is being removed, skipping TLS cleanup.")
+            return
+
+        self.charm.unit.status = MaintenanceStatus("Disabling TLS")
+
+        try:
+            self.charm._mysql.setup_client_tls()
+            self.charm._mysql.kill_client_sessions()
+        except MySQLTLSSetupError:
+            logger.error("Failed to disable TLS configuration.")
```
astrojuanlu: Same as above:


```suggestion
        except MySQLTLSSetupError as exc:
            logger.error(f"Failed to disable TLS configuration: {exc}")
```

machines/src/relations/tls.py
astrojuanlu: Or even `logger.exception` to show the full traceback

astrojuanlu (approved):
Left a few nits, leaving those to your discretion. Otherwise LGTM!

machines/src/relations/tls.py
sinclert-canonical: Done in commit https://github.com/canonical/mysql-operators/pull/248/commits/783b0b2d47fd6793dfe08fc63fcc5de470f13265.

machines/src/relations/tls.py
sinclert-canonical: Done in commit https://github.com/canonical/mysql-operators/pull/248/commits/783b0b2d47fd6793dfe08fc63fcc5de470f13265.

machines/src/relations/tls.py
sinclert-canonical: Done in commit https://github.com/canonical/mysql-operators/pull/248/commits/783b0b2d47fd6793dfe08fc63fcc5de470f13265.

kubernetes/src/charm.py
sinclert-canonical: I think we need to re-think how to make the access to the `self.charm.peers` property safer, to simplify a bunch of safe-related conditionals. Another option would be to remove all the SANs specification in the certificate request, as we are not yet leveraging the `VERIFY_IDENTITY` group-replication [SSL mode](https://dev.mysql.com/doc/refman/8.4/en/group-replication-secure-socket-layer-support-ssl.html) (only mode where the host-names / IPs are actually validated).

astrojuanlu: I think this accidentally introduced https://github.com/canonical/mysql-operators/issues/350 . Working on a fix.

PR #249, opened 2026-04-17
[MISC] 8.0 - Add K8s substrate to tutorial page

sinclert-canonical:
This PR adds the Kubernetes substrate to the documentation tutorial page.

The contents of the Kubernetes substrate have been extracted from commit `1cc36e1` and can be seen [here](https://github.com/canonical/mysql-operators/blob/1cc36e1a7577806ef66212c804e185db99bf7754/kubernetes/docs/tutorial/index.md).
This point in time is exactly one commit before the _unification of docs_ effort was merged.

#### Additional changes
- Replaced the usage of `root` user by `serverconfig`, similar to its `8.4/edge` docs counterpart.
- Fixed an issue with modern versions of `yq`, where keys containing dashes are not accessed unless quoted.
- Fixed an issue with how different substrates expose an IP. LXD's `public-address` versus K8s's `address`.

#### Review
See new [tutorial page](https://canonical-charmed-mysql--249.com.readthedocs.build/249/tutorial/).

PR #95, opened 2026-02-17
Merge VM and K8s documentation

a-velasco:
## Summary

The VM and K8s charms have near identical functionality, and ultimately represent the same product: Charmed MySQL. Having separate doc sets creates unnecessary barriers for contributions, maintenance, and general user experience of the charm.

This PR unifies the VM and K8s documentation pages located, respectively, at
* `/machines/docs`
* `/kubernetes/docs`

into a single set under
* `/docs`

This is done by merging all content that is identical, and using a combination of Sphinx tooling and restructuring to separate the differences between substrates (more details in the next section - [Differences between VM & K8s](#differences-between-vm-and-k8s)).

The only changes made to the technical content were:
* Fixing formatting and spelling errors that fail our automatic checks
* Updating links to the new repository where relevant
* Updating some headings to indicate VM or K8s-specific content
* Updating some easy and obvious inaccuracies like leftover `juju run-action` commands

A lot of pages have incorrect/inaccurate information, but this is carried over from the existing docs, and not an artifact of this PR. I've taken note of all the pages that require a technical review, and we can begin to address them once the unified docs are merged into `8.0/edge`.

## Differences between VM and K8s

Differences between documentation for VM and K8s charms were addressed in the following ways:

**Sphinx tabs**: This method was used in the vast majority of cases to separate substrate-specific information. The user's tab selection persists across other pages, so for example, a VM user will always see all tabs set to VM by default while navigating.

Example from `how-to/enable-tls.md`:
image

**Dedicated headings**: There is a small amount of pages that are tricky to unify due to how they are structured. Most of these pages require a refactor anyway, but for now, I thought this would be a viable solution.

Example from `how-to/deploy/airgapped.md`:
image

**Dedicated pages**: There are specific cases where entire pages are only relevant to one substrate at a time. Most notably, cloud deployment guides and release notes:

image

image

## Reviewing this PR

I don't believe it makes sense to heavily review the content of every single page of the docs, since this was not modified in major ways.

What _does_ need an engineering review in this PR are some specific pages where the VM/K8s separation was unclear.

- [x] [`How-to guides > Cluster-cluster replication > Deploy`](https://github.com/canonical/mysql-operators/blob/8.0/docs/docs/how-to/cluster-cluster-replication/deploy.md): Under the "Consume" section, the _Caution_ message regarding unit scaling only existed in the K8s version of this page. Does this warning also apply to VM?
[...]

PR #250, opened 2026-04-17
[MISC] 8.4 - Add K8s substrate to tutorial page

sinclert-canonical:
This PR adds the Kubernetes substrate to the documentation tutorial page.

The contents of the Kubernetes substrate have been extracted from commit `1cc36e1` and can be seen [here](https://github.com/canonical/mysql-operators/blob/1cc36e1a7577806ef66212c804e185db99bf7754/kubernetes/docs/tutorial/index.md).
This point in time is exactly one commit before the _unification of docs_ effort was merged.

#### Additional changes
- Replaced the usage of `root` user by `charmed-operator`, similar to its `8.0/edge` docs counterpart.
- Fixed an issue with modern versions of `yq`, where keys containing dashes are not accessed unless quoted.
- Fixed an issue with how different substrates expose an IP. LXD's `public-address` versus K8s's `address`.

#### Review
See new [tutorial page](https://canonical-charmed-mysql--250.com.readthedocs.build/250/tutorial/).

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -7,7 +7,7 @@ myst:
 (tutorial)=
 # Tutorial
 
-This hands-on tutorial aims to help you learn how to deploy Charmed MySQL on LXD / Kubernetes and become familiar with its available operations.
+This hands-on tutorial aims to help you learn how to deploy Charmed MySQL on LXD or Kubernetes and become familiar with its available operations. If you are not sure which path to take, we recommend following this tutorial on LXD by following the "VM" tabs.
 
 This hands-on tutorial will familiarize you with Charmed MySQL by going through a real deployment of a MySQL cluster. You will learn the basics of deploying, scaling, managing credentials, accessing a database, and other essential operations that will prepare you for more advanced deployments.
 
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -292,7 +292,7 @@ To retrieve the host’s IP address, run `juju show-unit`, and send the output t
 :user: ubuntu
 :host: my-vm
 
-DB_HOST=$(juju show-unit mysql/0 | yq '.[] | .address')
+DB_HOST=$(juju show-unit mysql/0 | yq '.[].["public-address"]')
 ```
 ````
 
@@ -303,7 +303,7 @@ DB_HOST=$(juju show-unit mysql/0 | yq '.[] | .address')
 :user: ubuntu
 :host: my-vm
 
-DB_HOST=$(juju show-unit mysql-k8s/0 | yq '.[] | .address')
+DB_HOST=$(juju show-unit mysql-k8s/0 | yq '.[].["address"]')
 ```
 ````
 `````
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -1,15 +1,15 @@
 ---
 myst:
   html_meta:
-    description: "Follow this hands-on tutorial to deploy, scale, and manage Charmed MySQL on virtual machines using Juju from setup to clean up."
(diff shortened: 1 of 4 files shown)


docs/tutorial.md
```diff
@@ -169,64 +246,74 @@ The easiest way to access MySQL is through the [MySQL Command-Line Client](https
...
-In a cluster with more units, **the primary is not necessarily equivalent to the leader**. To identify the primary unit in a cluster, run `juju run mysql/<any_unit> get-cluster-status`. 
-```
+### Interact with MySQL
 
-Once inside the Juju virtual machine, the `root` user can access MySQL with the following command:
+To access the unit hosting Charmed MySQL in production, one would normally use the following command:
 
 ```shell
-mysql -h 127.0.0.1 -uroot -p<password>
+mysql -h <ip_address> -u<username> -p<password>
```
paulomach: same as [8.0](https://github.com/canonical/mysql-operators/pull/249/changes#r3100216759)

paulomach (approved):

docs/tutorial.md
sinclert-canonical: Done in commit https://github.com/canonical/mysql-operators/pull/250/commits/8a6d390eda447c5fc212315977203a0149c4d32d.

docs/tutorial.md
```diff
@@ -1,15 +1,15 @@
...
   html_meta:
-    description: "Follow this hands-on tutorial to deploy, scale, and manage Charmed MySQL on virtual machines using Juju from setup to clean up."
+    description: "Follow this hands-on tutorial to deploy, scale, and manage Charmed MySQL using Juju from setup to clean up."
 ---
 
 (tutorial)=
 # Tutorial
 
-This hands-on tutorial aims to help you learn how to deploy Charmed MySQL on machines and become familiar with its available operations.
+This hands-on tutorial aims to help you learn how to deploy Charmed MySQL on LXD / Kubernetes and become familiar with its available operations.
```
a-velasco:
```suggestion
This hands-on tutorial aims to help you learn how to deploy Charmed MySQL on LXD or Kubernetes and become familiar with its available operations.

If you are not sure which path to take, we recommend following this tutorial on LXD by following the "VM" tabs. 
```

a-velasco (changes requested):
Thank you very much for these updates, @sinclert-canonical!

When unifying docs between substrates, we had decided to keep the tutorial VM-only since we want to be up and running in the simplest/most accessible way possible, and not give users too many choices. 

That being said, the K8s alternatives you added are rather un-intrusive and I can see how they might be a welcome alternative for some users. So let's apply this change, and get more feedback from users in the future to get a better sense of what they would like from a tutorial.

(Only change I'd like to make is clarifying that you should pick VM if you're not sure.)

sinclert-canonical:
> When unifying docs between substrates, we had decided to keep the tutorial VM-only since we want to be up and running in the simplest/most accessible way possible, and not give users too many choices.

I honestly did not recall. I proposed these changes as I saw that most of the docs are contain substrate specific snippets (even those referenced in the tutorial), but the tutorial itself. Let me know whether I should revert the content changes and leave the `public-address` vs `address` fix alone.

docs/tutorial.md
sinclert-canonical: Done in commit https://github.com/canonical/mysql-operators/pull/250/commits/d83218cedc880dd9425b7d50da24270c2637673c.

astrojuanlu:
> When unifying docs between substrates, we had decided to keep the tutorial VM-only since we want to be up and running in the simplest/most accessible way possible, and not give users too many choices.

The discussion was buried in a review comment https://github.com/canonical/mysql-operators/pull/95#discussion_r2821233326

a-velasco (approved):

PR #251, opened 2026-04-17
[TEST] test improvements

paulomach:


kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2370,13 +2371,16 @@ def force_quorum_from_instance(self) -> None:
...
             f"{self.cluster_admin_user}:{self.cluster_admin_password}@{self.instance_address}"
         )
+        # TODO: modify/expose timeout in the mysql-shell-client library
+        address = f"{instance_def}:3306"
+        command = "\n".join((
+            f"cluster = dba.get_cluster('{self.cluster_name}')",
+            f"cluster.force_quorum_using_partition_of('{address}')",
+        ))
 
+        executor = self._build_cluster_tcp_executor(self.instance_address)
```
paulomach, line 2380: @sinclert-canonical , testing this here, but how should we expose timeout at this level?

sinclert-canonical: At the moment, timeouts can only be applied in an operation-per-operation basis, in the `BaseExecutor` subclasses (see [code](https://github.com/canonical/mysql-shell-client/blob/9bbe9034ce5091176bbf75ae8170cc1d46439123/src/mysql_shell/executors/base.py#L28-L36)). That means that timeouts can only be applied when manually executing the Python / SQL operations via an executor class.

I originally thought about exposing this argument one level up the chain, in the client classes, but it feels noisy to do in an operation-per-operation basis (i.e. every class method will have an additional arg), and a bit overkill in their constructors, as not every operation should have the same timeout.

Open to ideas.

kubernetes/tests/integration/integration/high_availability/test_async_replication_upgrade.py
```diff
@@ -105,14 +107,23 @@ def test_build_and_deploy(first_model: str, second_model: str, charm: str) -> No
...
+    def wait_model_2():
+        model_2.wait(
+            ready=wait_for_apps_status(jubilant_backports.all_active, MYSQL_APP_2),
+            timeout=10 * MINUTE_SECS,
+        )
+
+    with ThreadPoolExecutor() as executor:
+        futures = [executor.submit(wait_model_1), executor.submit(wait_model_2)]
+        for future in futures:
+            future.result()
```
astrojuanlu, line 126: I'm curious, what does this accomplish? Practically speaking, the execution won't go past this point until both `.wait` calls return, which is what the sequential code does too.

kubernetes/tests/integration/integration/high_availability/test_async_replication_upgrade.py
```diff
@@ -157,16 +168,29 @@ def test_deploy_test_app(first_model: str) -> None:
         charm=MYSQL_TEST_APP_NAME,
         app=MYSQL_TEST_APP_NAME,
         base="ubuntu@22.04",
-        channel="latest/edge",
+        channel="latest/edge/racing",
         num_units=1,
         constraints=constraints,
     )
 
+    model_1.deploy(
```
astrojuanlu, line 176: Does this intersect with https://warthogs.atlassian.net/browse/DPE-9668 ?

kubernetes/lib/charms/mysql/v0/mysql.py
astrojuanlu: I understand that the upcoming mysql-shell-client 1.0 will have this?

paulomach:
no longer topical

PR #252, opened 2026-04-20
[MISC] Add UML diagram of high level architecture

astrojuanlu:
## Issue

Found myself lacking some visual summary of how the charm code is structured, so I propose including this UML diagram:

image

## Checklist

### DIFF
--- a/docs/explanation/architecture.md
+++ b/docs/explanation/architecture.md
@@ -197,6 +197,92 @@ Prometheus is an open-source systems monitoring and alerting toolkit with a dime
 
 See the charm state machines displayed in {ref}`flowcharts`. The low-level logic is mostly common for both VM and K8s charms.
 
+### Code architecture
+
+The charm codebase follows a layered architecture pattern that separates concerns
+between Juju charm orchestration and MySQL workload operations.
+This pattern is shared between both VM and K8s implementations:
+
+```{mermaid}
+classDiagram
+    direction TB
+
+    %% External framework
+    class ABC {
+        📄 abc.py (Python stdlib)
+    }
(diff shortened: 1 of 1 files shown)


sinclert-canonical (commented):
I am not sure this is going to help charm users rather than confuse them.

In principle our docs are mean to be user facing, and not developer facing. If you wonder why we have an _architecture_ titled section, or why we are listing all Juju events our charm react to (which is highly outdated), I do too. Those sections should probably be moved to some kind of hidden page only available to us, developers.

Will defer on @a-velasco and @paulomach for final approval.

astrojuanlu:
It's a good point. What's a good place to store dev-facing documentation? Should we enable the Wiki in this repository?

a-velasco:
@astrojuanlu The topic of developer docs has been kind of a long-standing discussion (that hasn't really reached a conclusion yet...)

Personally I think it makes sense to put docs that are very developer-specific, like event flowcharts and software architecture (at the class/function level) - in the GitHub Wiki (like how [OpenSearch](https://github.com/canonical/opensearch-operator/wiki) does it). What matters most is that it's easy for devs to find and contribute, imo. 

If we start populating the Wiki with more dev docs, we can always improve our signposting to it in the product docs, README, and CONTRIBUTING.

astrojuanlu:
Given that there are charms using it already, I went ahead and enabled the wiki for this repo, and moved the contents of this PR to https://github.com/canonical/mysql-operators/wiki

Happy to continue expanding that over time and adding pointers to it in the future as you suggest @a-velasco . For now, closing this PR.

PR #254, opened 2026-04-21
[MISC] 8.0 - Bump rolling-ops charmlib to v1

sinclert-canonical:
This PR bumps the `rolling-ops` charmlib to v1.

These changes are necessary to be able to react on multiple callbacks, thanks to the FIFO queue that v1 introduced. These callbacks will be used not only for restarting the workload (current use-case), but also to restart group replication whenever TLS gets enable / disable.

#### Compatibility

This PR **does not** change the peers relation name, which makes it **compatible** with MySQL 8.0 deployments.


astrojuanlu (approved):
Ignored the copy-pasted module, focused on the diff in charm code. LGTM!

paulomach (approved):

sinclert-canonical:
🗞️ Turns out **both** the `rollingops` v1 _charmlib_ and _Python package_ assume Python 3.12+, so we will not be able to bump the library version for MySQL 8.0 unless such constraint is lifted.

PR #255, opened 2026-04-21
[MISC] 8.4 - Bump rolling-ops charmlib to v1

sinclert-canonical:
This PR bumps the `rolling-ops` charmlib to v1.

These changes are necessary to be able to react on multiple callbacks, thanks to the FIFO queue that v1 introduced. These callbacks will be used not only for restarting the workload (current use-case), but also to restart group replication whenever TLS gets enable / disable.

#### Compatibility

This PR **does** change the peers relation name, which makes it **incompatible** with MySQL 8.4 deployments.

---

Depends on https://github.com/canonical/charmlibs/pull/419.


kubernetes/metadata.yaml
```diff
@@ -41,7 +41,7 @@ peers:
     interface: mysql_peers
   refresh-v-three:
     interface: refresh
-  restart:
+  rolling-ops:
```
sinclert-canonical, line 50: Does not make sense to call it `restart` anymore. That is just one use-case for using the rolling-ops lib.

machines/metadata.yaml
```diff
@@ -25,7 +25,7 @@ peers:
     interface: mysql_peers
   refresh-v-three:
     interface: refresh
-  restart:
+  rolling-ops:
```
sinclert-canonical, line 28: Does not make sense to call it `restart` anymore. That is just one use-case for using the rolling-ops lib.

kubernetes/src/charm.py
```diff
@@ -184,7 +184,12 @@ def __init__(self, *args):
             if self._refresh.workload_allowed_to_start:
                 self._refresh.next_unit_allowed_to_refresh = True
 
-        self.restart = RollingOpsManager(self, relation="restart", callback=self._restart)
+        # TODO: Use RollingOpsManager class when https://github.com/canonical/charmlibs/pull/419 gets merged
+        self.rolling_ops = PeerRollingOpsBackend(
```
astrojuanlu: I don't follow: if we're installing from `@ git+https://github.com/patriciareinoso/charmlibs@DPE-9350-etcd-lock-test`, can't we just use `RollingOpsManager` already?

astrojuanlu (commented):
Left 1 question

kubernetes/src/charm.py
sinclert-canonical: I am just iterating on that with Patricia.

The final goal is to use the `RollingOpsManager` class, as that is the one publicly available. However, the current Python package assumes a **mandatory** peer relation for a possible ETCD worker that coordinates inter-application rolling-ops operation (something required for MongoDB, but not for MySQL).

Therefore, I show-case the internal `PeerRollingOpsBackend` class as a way to circumvent that.

---

To be clear: we **must** migrate to the `RollingOpsManager` class before merging this PR. I am just waiting for the NoSQL folks to do the necessary changes. If they do not, we will be force to use the charmlib v1, similar to what the MySQL 8.0 [PR](https://github.com/canonical/mysql-operators/pull/254) proposes.

paulomach (approved):
Ok - but we wait on the merge

kubernetes/src/charm.py
sinclert-canonical: Just migrated to the `RollingOpsManager` class.

sinclert-canonical:
🗞️ This PR has been updated to validate its compatibility with curren rolling-ops v1 Python package efforts:

- Created + granted container user access to logs directory (`/var/lib/rollingops`).
- Migrated to the intended rolling-ops v1 manager class (`RollingOpsManager`).
- Rebased branch with respect to `8.4/edge`.

---

Up to @paulomach how much do we want to wait for https://github.com/canonical/charmlibs/pull/445 to get merged.

astrojuanlu:
I'd also be in favour of waiting for https://github.com/canonical/charmlibs/pull/445

astrojuanlu (approved):
Other than that, LGTM!

sinclert-canonical:
PR branch rebased from the `8.4/edge` branch.

PR #87, opened 2026-02-13
[DPE-7908] Separation of storage

astrojuanlu:
## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.

PR #256, opened 2026-04-21
[MISC] Miscellaneous cleanups

astrojuanlu:
## Issue

Ended up accumulating lots of cleanups in #87, so I decided to send a separate pull request:

_**Note**: This does _not_ include adapting to the new backup helpers, because it turns out those need a writable temporary directory. Instead of introducing some interim hack, I prefer to do both adaptation + separation of storage in a single PR._

Main changes:
- Fix backup methods in VM and K8s child classes so that they have a compatible signature with the parent class
- Fix type hints of TLS handling code (see also https://github.com/canonical/object-storage-integrator/pull/170)
- Visually separate TLS handling in backups code
  - This introduces some repeated error handling, with the benefit of improved readability
- Make Ceph backup test amenable for local development
  -  https://github.com/canonical/mysql-operators/pull/88 had introduced some changes that were not properly accounted for in this test
- Improved support for ty
  - Including some ty configurations in `pyproject.toml` but also some actual typing improvements
- Minor improvements in observability, metadata, typos

## Checklist

### DIFF
--- a/kubernetes/lib/charms/mysql/v0/backups.py
+++ b/kubernetes/lib/charms/mysql/v0/backups.py
@@ -118,7 +118,7 @@ MOVE_RESTORED_CLUSTER_TO_ANOTHER_S3_REPOSITORY_ERROR = (
 )
 
 if typing.TYPE_CHECKING:
-    from mysql import MySQLCharmBase
+    from .mysql import MySQLCharmBase
 
 
 class MySQLBackups(Object):
--- a/kubernetes/lib/charms/mysql/v0/backups.py
+++ b/kubernetes/lib/charms/mysql/v0/backups.py
@@ -727,7 +727,7 @@ class MySQLBackups(Object):
             self.charm._mysql.delete_temp_backup_directory()
             # Old backups may contain the temp backup directory (as previously, the temp
             # backup directory was created in the mysql data directory to reduce IOPS latency)
-            self.charm._mysql.delete_temp_backup_directory(from_directory=MYSQL_DATA_DIR)
+            self.charm._mysql.delete_temp_backup_directory(tmp_base_directory=MYSQL_DATA_DIR)
         except MySQLDeleteTempRestoreDirectoryError:
             return False, "Failed to delete the temp restore directory"
         except MySQLDeleteTempBackupDirectoryError:
--- a/kubernetes/lib/charms/mysql/v0/backups.py
+++ b/kubernetes/lib/charms/mysql/v0/backups.py
@@ -265,7 +265,10 @@ class MySQLBackups(Object):
             return False
 
         if "s3-block-message" in self.charm.app_peer_data:
-            logger.error("Backup failed: S3 relation is blocked for write")
+            logger.error(
+                "Backup failed: S3 relation is blocked for write: %s",
+                self.charm.app_peer_data["s3-block-message"],
+            )
             event.fail("S3 relation is blocked for write")
             return False
 
--- a/kubernetes/lib/charms/mysql/v0/mysql.py
+++ b/kubernetes/lib/charms/mysql/v0/mysql.py
@@ -2475,7 +2475,7 @@ class MySQLBase(ABC):
     def execute_backup_commands(
         self,
         s3_path: str,
-        s3_parameters: dict[str, str],
+        s3_parameters: dict,
         xtrabackup_location: str,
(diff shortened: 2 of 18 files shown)


astrojuanlu:
Upgrade tests failing with

```
unit-mysql-0: 21:20:34 INFO unit.mysql/0.juju-log charm_refresh:Refresh incompatible because new version of MySQL ('8.4') and/or charm (CharmVersion('8.4/1.22.0.post13.dev0+6d221716')) is not compatible with previous version of MySQL ('8.4') and/or charm (CharmVersion('8.4/1.15.0'))
```

The rest seem just flaky or slow, I'll restart them a few more times.

paulomach (approved):
Nothing polemic, better optic

kubernetes/lib/charms/mysql/v0/backups.py
```diff
@@ -118,7 +118,7 @@ def is_unit_blocked(self) -> bool:
 )
 
 if typing.TYPE_CHECKING:
-    from mysql import MySQLCharmBase
+    from .mysql import MySQLCharmBase
```
sinclert-canonical, line 121: AFAIK, relative imports cannot be done outside of a _package_ (i.e. a folder with a `__init__.py` file in it).
Am I missing something?

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2448,7 +2448,7 @@ def get_available_memory(self) -> int:
         """Platform dependent method to get the available memory for mysql-server."""
         raise NotImplementedError
 
-    def create_sa_pem_file(
+    def create_ca_pem_file(
```
sinclert-canonical, line 2451: Thanks, it gets to my nerves as well.

kubernetes/pyproject.toml
```diff
@@ -144,3 +144,6 @@ max-complexity = 10
 
 [tool.ruff.lint.pydocstyle]
 convention = "google"
+
+[tool.ty.environment]
+extra-paths = ["./lib"]
```
sinclert-canonical: I understand you want to use `ty` as our type checker, but AFAIK there has been no consensus on what tool will all Data Platform adopt (unlike `tox`, `poetry` and `ruff`).

This should be removed.

sinclert-canonical (commented):
I see we are not fully decoupling the generation of the _CA chain location_ from the creation of the _backup temporal directory_, as discussed online. I assume there is a good reason for this, because last time we spoke we were on the same page about splitting them.

The proposed approach is alright to me.

---

The same comments left for the K8s implementation, also apply to the VM one.

kubernetes/pyproject.toml
astrojuanlu: Fair enough, I'll see how can I keep the config locally

kubernetes/lib/charms/mysql/v0/backups.py
astrojuanlu: That's also what I thought. It seems to do the trick, probably because the IDE or the type checkers are being too permissive.

I'd actually rather change these imports to absolute ones, but didn't want to pollute this PR even more.

astrojuanlu:
> I see we are not fully decoupling the generation of the CA chain location from the creation of the backup temporal directory, as discussed online. I assume there is a good reason for this, because last time we spoke we were on the same page about splitting them.

Yes. On one hand, the CA chain file creation is already split:

https://github.com/canonical/mysql-operators/blob/e10ede0bd41847ffd87923393075f6172d8c2c10/machines/lib/charms/mysql/v0/mysql.py#L2451

And on the other hand, I was misguided - I thought the CA chain file creation was the _exception_, but it wasn't: executing the wrapper as `root` was. More rationale in the description of #257, happy to elaborate more or explain synchronously if needed.

kubernetes/lib/charms/mysql/v0/backups.py
sinclert-canonical: > It seems to do the trick, probably because the IDE or the type checkers are being too permissive.

Which trick? AFAIK these imports have been working until today. Did `ty` complaint about them?

sinclert-canonical (approved):

astrojuanlu:
Tests were passing before, so I'm merging and rebasing #257.

PR #247, opened 2026-04-16
[MISC] Remove permissions hack from backups using wrapped apps (VM)

astrojuanlu:
## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.

PR #257, opened 2026-04-21
[DPE-7908] Separation of storage (take 2)

astrojuanlu:
## Issue

(Depends on #256)

I went back and forth over several days about this. There's a series of unfortunate circumstances that make it difficult to find an elegant, simple implementation.

- One cannot execute the backup scripts as `snap_daemon`. This user has the home directory set to `/nonexistent`, and any attempt to trigger it (`sudo -u`, `subprocess.Popen(user=)`, `runuser -u`) will trigger a snap confinement error.
- As such, a different approach is needed to avoid running the backup scripts as `root`. Specifically, https://github.com/canonical/charmed-mysql-snap/pull/104. This uses `setpriv` as recommended in [the official Snap documentation](https://snapcraft.io/docs/explanation/snap-development/system-usernames/#dropping-privileges) and as done for other apps already.
- _However_, only `root` can run `setpriv`, which means that the wrapped scripts must all run as `root` to drop privileges properly.
- And yet _everything else_ that the wrapped scripts themselves need must be readable by `snap_daemon`. This includes
  - The CA file, and
  - Any temporary directories
- Initially I thought that creating the CA file as `snap_daemon` and let everything else use `user` and `group` would suffice. But this was wrong. The backup methods require creating some temporary directories in writable locations.
- As such, everything must run as `snap_daemon`, except the scripts themselves, which must be run as `root`.
- Similarly to my initial attempt, these were the options I came up with:
  1. Copy-paste all the logic in the machines subclass, tweaking what I need. Felt excessive for changing the argument of 1 function.
  2. Introduce some extra parameter in the lib methods that must be passed around for the child class to detect when to force `root`. Seemed similar to what I initially tried in #247.
  3. Special-case the wrapped scripts in `_execute_commands` in the child subclass. Ugly, but effective. This is what's implemented here.
  4. Refactor the backup methods to introduce private methods that would run the inner script which could be overridden by the child classes. Something like

```python
    def execute_backup_commands(self, ...):
        # pre-backup checks
        self._do_execute_backup_commands(self, ...)  # Overridden by the child classes
```

However, I think this might introduce a lot of indirection, some duplicity, and little benefit.

---

Ultimately, the tension comes because the `user` parameter clashes with the need to execute the wrappers as `root`. The solution is clear, but the architecture of the code makes it hard to actually implement.

## Checklist


paulomach:
Thanks for the context. I had happily forgot all these details.

kubernetes/src/mysql_k8s_helpers.py
```diff
@@ -577,8 +624,8 @@ def is_data_dir_initialised(self) -> bool:
 
             # minimal expected content for an integral mysqld data-dir
             expected_content = {
-                "#innodb_redo",
-                "#innodb_temp",
+                # "#innodb_redo",  # stored separately
+                # "#innodb_temp",  # stored separately
```
paulomach: Remove comment?

machines/src/mysql_vm_helpers.py
```diff
@@ -599,6 +606,19 @@ def _execute_commands(
...
+            for wrapped_command in (
+                CHARMED_MYSQL_XTRABACKUP_LOCATION,
+                CHARMED_MYSQL_XBCLOUD_LOCATION,
+                CHARMED_MYSQL_XBSTREAM_LOCATION,
+            )
+        ):
+            # Wrapped commands must be run as root because they use setpriv to drop privileges,
+            # see https://snapcraft.io/docs/explanation/snap-development/system-usernames/#dropping-privileges
+            logging.info("Running wrapped command as root: %s", commands)
+            user, group = ROOT_SYSTEM_USER, ROOT_SYSTEM_USER
```
sinclert-canonical, line 620: I guess this is the gist of it: the new way of dealing with `snap_daemon` / `root` required privileges in a god-like function (i.e. a function run to perform many different operations).

Great job encapsulating all of it in a single conditional.

sinclert-canonical (approved):
I reviewed commits https://github.com/canonical/mysql-operators/pull/257/changes/52e4a2c4ba34586dcabbd0e545cb8e44e7a598c8, https://github.com/canonical/mysql-operators/pull/257/changes/8e50a82a24c3a1e1f7c4d17429553808c2171615 and https://github.com/canonical/mysql-operators/pull/257/commits/5dc63571c29ff90e5c47f0bf5b55d2a71ccf2496, which I assumed are the ones built on top of PR #256.

They are exactly the same as the ones in the previously approved _separation of storage PR_, aside from the encapsulation of the _required root user_ switch in a single conditional (which I find highly convenient).

Fantastic investigation 💯

machines/src/mysql_vm_helpers.py
```diff
@@ -872,39 +896,48 @@ def read_file_content(self, path: str) -> str | None:
...
+        """Remove all files from the data directory."""
         logger.warning(f"Resetting data directory: {MYSQL_DATA_DIR}")
 
-        # Remove the data directory
-        shutil.rmtree(MYSQL_DATA_DIR, ignore_errors=False)
-
-        # Recreate the data directory
-        os.makedirs(MYSQL_DATA_DIR)
+        # Remove the contents of the data directory
+        for root, dirs, files in pathlib.Path(MYSQL_DATA_DIR).walk(top_down=False):
```
paulomach, line 910: til `Path.walk`, introduced in 3.12 so good for noble onward

kubernetes/src/mysql_k8s_helpers.py
```diff
@@ -334,8 +350,8 @@ def retrieve_backup_with_xbcloud(
         temp_restore_directory: str = MYSQL_DATA_DIR,
```
paulomach: Should we use the temp dir here also?

kubernetes/src/mysql_k8s_helpers.py
paulomach: Should we use the temp dir here also?

kubernetes/src/mysql_k8s_helpers.py
```diff
@@ -352,42 +368,73 @@ def retrieve_backup_with_xbcloud(
...
+            mysql_data_directory,
+            xtrabackup_plugin_directory,
+            user,
+            group,
         )
 
-    def delete_temp_restore_directory(self) -> None:
+    def delete_temp_restore_directory(
+        self,
+        temp_restore_directory: str = MYSQL_DATA_DIR,
```
paulomach: Should we use the temp dir here also?

paulomach (commented):
Left some questions - otherwise is good given the constraints

kubernetes/src/mysql_k8s_helpers.py
sinclert-canonical: +1

kubernetes/src/mysql_k8s_helpers.py
astrojuanlu: Indeed, will change

kubernetes/src/mysql_k8s_helpers.py
astrojuanlu: Indeed, will change

kubernetes/src/mysql_k8s_helpers.py
astrojuanlu: Indeed, will change

astrojuanlu:
Some stuff failing on K8s, looking into it.

astrojuanlu:
After closer inspection, I figured that the failing tests are probably not related to this PR. So I'm merging as-is.

astrojuanlu:
Proposal for self-healing test that was failing here https://github.com/canonical/mysql-operators/pull/264

paulomach (approved):

astrojuanlu:
There are some test failures from the usual suspects. I'm going ahead but I'll keep an eye on these tests for the coming days.

PR #258, opened 2026-04-22
[DPE-9788] Move Allure results to subdirectory depending on branch

astrojuanlu:
## Checklist

### DIFF
--- a/.github/workflows/schedule.yaml
+++ b/.github/workflows/schedule.yaml
@@ -89,6 +89,10 @@ jobs:
         with:
           ref: gh-pages-beta
           path: repo/
+      - name: Resolve report directory from branch name
+        # ALLURE_REPORT_DIR will be 8_4_edge for 8.4/edge, 8_0_edge for 8.0/edge and so forth
+        run: |
+          echo "ALLURE_REPORT_DIR=$(echo \"${GITHUB_REF_NAME}\" | tr '[:punct:]' '_')" >> "${GITHUB_ENV}"
       - name: Download default test results
         # Default test results in case the integration tests time out or runner set up fails
         # (So that Allure report will show "unknown"/"failed" test result, instead of omitting the test)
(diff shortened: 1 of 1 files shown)


paulomach (approved):
How to test this?

astrojuanlu:
Working this out on my fork, ⌛

.github/workflows/schedule.yaml
```diff
@@ -89,6 +89,10 @@ jobs:
         with:
           ref: gh-pages-beta
           path: repo/
+      - name: Resolve report directory from branch name
+        # ALLURE_REPORT_DIR will be 8_4_edge for 8.4/edge, 8_0_edge for 8.0/edge and so forth
```
sinclert-canonical: Nit: please move the comment within the `run` block, directly above of the line it targets.

.github/workflows/schedule.yaml
```diff
@@ -89,6 +89,10 @@ jobs:
...
+        run: |
+          echo "ALLURE_REPORT_DIR=$(echo \"${GITHUB_REF_NAME}\" | tr '[:punct:]' '_')" >> "${GITHUB_ENV}"
```
sinclert-canonical: This is not exactly the same as how PostgreSQL is doing it (see [variable](https://github.com/canonical/postgresql-operator/blob/16730dcf84d910e73ad5c9f6eb4b58a8875398de/.github/workflows/integration_test.yaml#L402)). Have you tried their approach? Do we need to deviate for some reason?

astrojuanlu: No particular reason to be fair. Our approach to fill the variable is a one-liner, compared to https://github.com/canonical/postgresql-operator/blob/16730dcf84d910e73ad5c9f6eb4b58a8875398de/.github/workflows/integration_test.yaml#L226-L239 . But theirs is probably safer. I'll port that one to ours and match the variable name.

sinclert-canonical: Ok. No need to move everything to the `environment` section just yet. I just noticed that they are using GitHub provided `github.run_number` value as the variable for the GH pages branch.

I think that is the part we should copy.

astrojuanlu:
Test workflow running on my fork https://github.com/astrojuanlu/mysql-operators-debug/actions/runs/24825811796 with the same code https://github.com/astrojuanlu/mysql-operators-debug/blob/86ff7e61039d8839dafea488f982c572227311a0/.github/workflows/schedule.yaml#L62

.github/workflows/schedule.yaml
astrojuanlu: Ended up bring this closer to the postgres-operator method

sinclert-canonical (approved):

astrojuanlu:
This is only touching the `schedule.yaml` workflow, so I'm merging as soon as #259 is approved.

PR #261, opened 2026-04-22
[MISC] 8.4 - Bump pydantic library to v2

sinclert-canonical:
This PR bumps the `pydantic` library to v2.

These changes are necessary to be able to install the upcoming `rolling-ops` Python package (see [PR](https://github.com/canonical/charmlibs/pull/419)), which will depend on this specific version of the library. In order to do so, two of the Data Platform libs needed to be bumped.

#### Compatibility

This PR **does not** change functionality, which makes it incompatible with MySQL 8.4 deployments.

---

Unblocks https://github.com/canonical/mysql-operators/pull/255.

### DIFF
--- a/kubernetes/lib/charms/data_platform_libs/v1/data_models.py
+++ b/kubernetes/lib/charms/data_platform_libs/v1/data_models.py
@@ -1,4 +1,4 @@
-# Copyright 2023 Canonical Ltd.
+# Copyright 2024 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -39,7 +39,7 @@ also parsing and validation on standard dataclass implementation:
 
 ```python
 
-from charms.data_platform_libs.v0.data_models import BaseConfigModel
+from charms.data_platform_libs.v1.data_models import BaseConfigModel
 
 class MyConfig(BaseConfigModel):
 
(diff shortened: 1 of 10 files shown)


paulomach (approved):

kubernetes/lib/charms/data_platform_libs/v1/data_models.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2023 Canonical Ltd.
+# Copyright 2024 Canonical Ltd.
```
astrojuanlu, line 1: 🤔  https://legalclarity.org/should-the-copyright-year-be-updated/

> A standard copyright notice includes three specific parts:
> 
> The symbol © (the letter C in a circle), the word “Copyright,” or the abbreviation “Copr.”
> The year the work was **first published**.

(Emphasis mine)

kubernetes/src/config.py
```diff
@@ -51,18 +51,18 @@ class CharmConfig(BaseConfigModel):
...
 
     profile: str
-    cluster_name: str | None
-    cluster_set_name: str | None
-    profile_limit_memory: int | None
-    experimental_max_connections: int | None
+    cluster_name: str | None = Field(default=None)
+    cluster_set_name: str | None = Field(default=None)
+    profile_limit_memory: int | None = Field(default=None)
+    experimental_max_connections: int | None = Field(default=None)
```
astrojuanlu, line 57: nit:

I've always been a fan of the [`Annotated` pattern](https://pydantic.dev/docs/validation/latest/concepts/fields/#the-annotated-pattern). The alternative (implemented here) amounts to

```
cluster_name: str = Field(...)
```

but technically `Field` _is not_ a `str`. This always irked me, hence the suggestion.

For an example of another DP team applying it, see https://github.com/canonical/object-storage-integrator/blob/c8e4a827e1ab94bf9fb890bce50a75fc66dc0b61/s3/src/core/domain.py#L89-L121 (although to be fair they don't do it consistently)

astrojuanlu (approved):
Approved with 2 nits!

kubernetes/lib/charms/data_platform_libs/v1/data_models.py
sinclert-canonical: This comment comes from the shared library, in short: I copy+pasted the whole module.

However, I do remember that this work was first drafted in 2024 (see [PR](https://github.com/canonical/data-platform-libs/pull/201)), and later delayed in a never-ending _we do not actually need it_ process until PostgreSQL team needed it in May 2025... If you want to laugh even harder, take a look at the crossed out ~upgrade~ module from the PR description, and how, two years later, the NoSQL team needed to publish a v1 of that module as well (see [PR comment](https://github.com/canonical/data-platform-libs/pull/260#pullrequestreview-3889677747)).

_Anticipating future needs since 2024_ ✌🏻

kubernetes/src/config.py
sinclert-canonical: I am not sure how we could simplify code using the _Annotated_ type hint. I tried using it just in case we could save the `Field(default=None)` part, but that does not seem to be the case (unit tests do not pass).

PR #264, opened 2026-04-23
[MISC] Fix wrong waiting logic in self-healing test, undo Juju 2 workarounds

astrojuanlu:
## Issue

This test has always been broken, see https://canonical.github.io/mysql-operators/89/#suites/39f0662952b7076f02d119b4ba71c27e/47b6f482785863e3/history

image

With these changes, it passes locally.

## Checklist

### DIFF
--- a/kubernetes/tests/integration/integration/high_availability/test_self_healing_setup_crash.py
+++ b/kubernetes/tests/integration/integration/high_availability/test_self_healing_setup_crash.py
@@ -2,7 +2,6 @@
 # See LICENSE file for licensing details.
 
 import logging
-import os
 
 import jubilant
 from jubilant import Juju
@@ -11,17 +10,16 @@ from ...helpers_ha import (
     CHARM_METADATA,
     delete_k8s_pod,
     get_mysql_primary_unit,
-    load_mysql_test_data,
     update_interval,
     wait_for_apps_status,
-    wait_for_unit_status,
+    wait_for_unit_message,
 )
 
 MYSQL_APP_NAME = "mysql-k8s"
 MINUTE_SECS = 60
 
 
-def test_deploy_single_unit_cluster(juju: Juju, charm: str) -> None:
+def test_deploy_and_crash_during_cluster_setup_and_recover(juju: Juju, charm: str) -> None:
     """Simple test to ensure that the MySQL and application charms get deployed."""
     logging.info("Deploying MySQL cluster")
     juju.deploy(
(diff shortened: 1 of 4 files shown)


astrojuanlu:
Oh btw, months ago I introduced some subtle bugs in our tests... pushed 1 more commit that fixes those.

astrojuanlu:
For full context, @sinclert-canonical pointed out that he had already tried #229 + #230, but tests weren't passing on 8.0/edge. I contend the `successes=1` is the key part here, it decreases the probability of a race condition.

Test passing on 8.4 (this PR):
- https://github.com/canonical/mysql-operators/actions/runs/24832514612/job/72686596148?pr=264
- https://github.com/canonical/mysql-operators/actions/runs/24832514612/job/72686595961?pr=264

And on 8.0 (#265):
- https://github.com/canonical/mysql-operators/actions/runs/24838979947/job/72712778128?pr=265
- https://github.com/canonical/mysql-operators/actions/runs/24838979947/job/72712777722?pr=265
- https://github.com/canonical/mysql-operators/actions/runs/24838979947/job/72712778138?pr=265

So I think this and the companion PR are safe to merge.

kubernetes/tests/integration/integration/high_availability/test_self_healing_setup_crash.py
```diff
@@ -51,12 +53,21 @@ def test_crash_during_cluster_setup(juju: Juju, charm: str) -> None:
     """
     mysql_primary = get_mysql_primary_unit(juju, MYSQL_APP_NAME)
 
+    # NOTE: This is prone to race conditions:
+    # if the units clear the "waiting" phase too quickly,
+    # this status function will never activate
```
sinclert-canonical: No need to leave a comment if we are going to fix the problem here and now.

kubernetes/tests/integration/integration/high_availability/test_self_healing_setup_crash.py
```diff
@@ -51,12 +53,21 @@ def test_crash_during_cluster_setup(juju: Juju, charm: str) -> None:
...
     logging.info("Scaling to 3 units")
     juju.add_unit(MYSQL_APP_NAME, num_units=2)
     juju.wait(
-        ready=wait_for_apps_status(jubilant.any_waiting, MYSQL_APP_NAME),
+        ready=lambda status: any((
+            *(
+                wait_for_unit_status(MYSQL_APP_NAME, unit_name, "waiting")(status)
+                for unit_name in status.get_units(MYSQL_APP_NAME)
+            ),
+        )),
```
sinclert-canonical, line 42: I am not sure having _any_ unit in the `waiting` status is the required condition for this test to validate behavior upon _cluster setup failure_. There is a reason why this wait condition was initially targeting the application: we need to make sure no unit has actually setup the cluster.

Therefore, I think there are 3 possible ways to achieve this:
- A) Wait for `maintenance` status at the app level.
- B) Wait for `maintenance` status in **all** the units.
- C) Wait for `waiting` status in **all** the units.

Given how brief (+ juju-controller dependent) the `waiting` status is, I would argue A or B are best.

sinclert-canonical (commented):
Thanks for fixing the role tests!

Could you try reverting those tests wait conditions, to check for the app status, to see if it works? AFAIK, the approach was changed in this [PR](https://github.com/canonical/mysql-k8s-operator/pull/699) because of a Juju issue. Maybe that got solved 🤷🏻‍♂️

astrojuanlu:
https://github.com/juju/juju/issues/22307 makes it more difficult to reliably debug this issue. Moreover, milliseconds arent' shown...

astrojuanlu:
@paulomach suggested deploying 3 units directly, killing the primary as soon as it's configured, and waiting for the cluster to recover. That's simpler to reason about and free of race conditions.

astrojuanlu:
And not load any test data for convenience.

kubernetes/tests/integration/integration/high_availability/test_self_healing_setup_crash.py
astrojuanlu: Changed strategy, see https://github.com/canonical/mysql-operators/pull/264#issuecomment-4328638542

kubernetes/tests/integration/integration/high_availability/test_self_healing_setup_crash.py
```diff
@@ -11,16 +10,16 @@
...
     wait_for_apps_status,
+    wait_for_unit_message,
 )
 
 MYSQL_APP_NAME = "mysql-k8s"
 MINUTE_SECS = 60
 
 
-def test_deploy_single_unit_cluster(juju: Juju, charm: str) -> None:
+def test_deploy_and_crash_during_cluster_setup_and_recover(juju: Juju, charm: str) -> None:
```
sinclert-canonical: Please keep both "tests" split, instead of having a single big one where we deploy + do the checks.

All other HA tests follow this pattern.

sinclert-canonical:
> Could you try reverting those tests wait conditions, to check for the app status, to see if it works? AFAIK, the approach was changed in this https://github.com/canonical/mysql-k8s-operator/pull/699 because of a Juju issue. Maybe that got solved

Did you try this? Are tests checking for application `active` status + idle agents still failing?

kubernetes/tests/integration/integration/high_availability/test_self_healing_setup_crash.py
astrojuanlu: Fair enough. I did it this way because the deployment is "incomplete" (we don't wait until the app settles).

astrojuanlu:
> Did you try this? Are tests checking for application active status + idle agents still failing?

Oh I see what you mean, those idioms were introduced before we branched out 8.4/edge and dropped support for Juju 2.9. Fixed.

sinclert-canonical (approved):

astrojuanlu:
Rebased on top of current 8.4/edge, @paulomach

paulomach (approved):

PR #271, opened 2026-04-27
[DPE-7899] 8.4 - Migrate to TLS v4 (II)

sinclert-canonical:
This PR exposes a way to encrypt _peer-to-peer_ traffic using version 4 of the `tls-certificates` interface.

### Approach:

According to the docs (see last paragraph of [this page](https://dev.mysql.com/doc/refman/8.4/en/group-replication-secure-socket-layer-support-ssl.html)), when MySQL instances are clustered using `mysql` communication stack, the same variables that encrypt group-replication recovery, are used for group-replication communication. Therefore, [this page](https://dev.mysql.com/doc/refman/8.4/en/group-replication-configuring-ssl-for-recovery.html) specifies the variables used to configure the group-replication encryption.

In addition, MySQL group-replication encryption allows to define a _mode_. From those, the `REQUIRED` mode is proposed as a way to simplify the dynamic enabling / disabling of peer-to-peer encryption, in clusters whose instances will default to their own auto-generated certificates as a fallback (i.e. certificates coming from different CAs).

### Future work:

- Define the necessary logic to parse a manually configured private key for TLS certificates generation.

---

Depends on PR https://github.com/canonical/mysql-operators/pull/255

### DIFF
--- a/kubernetes/lib/charms/mysql/v0/mysql.py
+++ b/kubernetes/lib/charms/mysql/v0/mysql.py
@@ -2921,15 +2921,46 @@ class MySQLBase(ABC):
         require_tls: bool = False,
     ) -> None:
         """Setup client-connection TLS files and requirement mode."""
-        tls_var = "require_secure_transport"
-        tls_val = "ON" if require_tls else "OFF"
+        client = self._instance_client_tcp
+        usage = "ON" if require_tls else "OFF"
 
         try:
-            self._instance_client_tcp.set_instance_variable(Scope.PERSIST, "ssl_ca", ca_path)
-            self._instance_client_tcp.set_instance_variable(Scope.PERSIST, "ssl_key", key_path)
-            self._instance_client_tcp.set_instance_variable(Scope.PERSIST, "ssl_cert", cert_path)
-            self._instance_client_tcp.set_instance_variable(Scope.PERSIST, tls_var, tls_val)
-            self._instance_client_tcp.reload_instance_certs()
+            client.set_instance_variable(Scope.PERSIST, "ssl_ca", ca_path)
+            client.set_instance_variable(Scope.PERSIST, "ssl_key", key_path)
+            client.set_instance_variable(Scope.PERSIST, "ssl_cert", cert_path)
+            client.set_instance_variable(Scope.PERSIST, "require_secure_transport", usage)
+            client.reload_instance_certs()
+        except ExecutionError as e:
+            raise MySQLTLSSetupError() from e
+
+    def setup_group_tls(
+        self,
+        ca_path: str = "ca.pem",
+        key_path: str = "server-key.pem",
+        cert_path: str = "server-cert.pem",
+        require_tls: bool = False,
+    ) -> None:
+        """Setup group-replication TLS files and requirement mode."""
+        client = self._instance_client_tcp
+        usage_recovery = "ON" if require_tls else "OFF"
+        usage_replication = "REQUIRED" if require_tls else "DISABLED"
+
+        try:
+            client.set_instance_variable(
+                Scope.PERSIST, "group_replication_recovery_ssl_ca", ca_path
+            )
+            client.set_instance_variable(
+                Scope.PERSIST, "group_replication_recovery_ssl_key", key_path
+            )
+            client.set_instance_variable(
+                Scope.PERSIST, "group_replication_recovery_ssl_cert", cert_path
+            )
+            client.set_instance_variable(
+                Scope.PERSIST, "group_replication_recovery_use_ssl", usage_recovery
+            )
+            client.set_instance_variable(
+                Scope.PERSIST, "group_replication_ssl_mode", usage_replication
+            )
         except ExecutionError as e:
             raise MySQLTLSSetupError() from e
 
--- a/kubernetes/tests/integration/integration/test_tls.py
+++ b/kubernetes/tests/integration/integration/test_tls.py
@@ -3,7 +3,6 @@
 
 import json
 import logging
-from time import sleep
 
 import jubilant
 from jubilant import Juju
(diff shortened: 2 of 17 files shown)


kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2921,15 +2921,36 @@ def setup_client_tls(
...
+    def setup_group_tls(
+        self,
+        ca_path: str = "ca.pem",
+        key_path: str = "server-key.pem",
+        cert_path: str = "server-cert.pem",
+        require_tls: bool = False,
+    ) -> None:
+        """Setup group-replication TLS files and requirement mode."""
+        client = self._instance_client_tcp
+        prefix = "group_replication"
```
paulomach: is `prefix` just to save line length?

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2921,15 +2921,36 @@ def setup_client_tls(
...
+        usage = "ON" if require_tls else "OFF"
+
+        try:
+            client.set_instance_variable(Scope.PERSIST, f"{prefix}_recovery_ssl_ca", ca_path)
+            client.set_instance_variable(Scope.PERSIST, f"{prefix}_recovery_ssl_key", key_path)
+            client.set_instance_variable(Scope.PERSIST, f"{prefix}_recovery_ssl_cert", cert_path)
+            client.set_instance_variable(Scope.PERSIST, f"{prefix}_recovery_use_ssl", usage)
+            client.set_instance_variable(Scope.PERSIST, f"{prefix}_ssl_mode", "REQUIRED")
```
paulomach: If `usage` is `OFF`, how the `REQUIRED` for `group_replication_ssl_mode` plays out?
Should we also set `group_replication_recovery_use_ssl` to hardcoded `ON`

kubernetes/tests/integration/integration/test_tls.py
```diff
@@ -104,9 +97,16 @@ def test_enable_tls(juju: Juju) -> None:
...
     juju.integrate(f"{APP_NAME}:client-certificates", f"{TLS_APP_NAME}:certificates")
+    juju.wait(
+        ready=wait_for_apps_status(jubilant.all_active, TLS_APP_NAME),
+        timeout=TIMEOUT,
+    )
 
-    # allow time for TLS enablement
-    sleep(TLS_SETUP_SLEEP_TIME)
+    juju.integrate(f"{APP_NAME}:peer-certificates", f"{TLS_APP_NAME}:certificates")
+    juju.wait(
```
paulomach, line 107: I agree removing sleep is good practice, but how are we ensuring that MySQL was reconfigured in time before the assertions?

kubernetes/tests/integration/integration/test_tls.py
```diff
@@ -132,9 +132,16 @@ def test_disable_tls(juju: Juju) -> None:
...
     logger.info("Removing relation")
     juju.remove_relation(f"{APP_NAME}:client-certificates", f"{TLS_APP_NAME}:certificates")
+    juju.wait(
+        ready=wait_for_apps_status(jubilant.all_active, TLS_APP_NAME),
+        timeout=TIMEOUT,
+    )
 
-    # Allow time for reconfigure
-    sleep(TLS_SETUP_SLEEP_TIME)
+    juju.remove_relation(f"{APP_NAME}:peer-certificates", f"{TLS_APP_NAME}:certificates")
```
paulomach, line 143: I agree removing sleep is good practice, but how are we ensure that MySQL was reconfigured in time before the assertions?

kubernetes/tests/integration/integration/test_tls.py
paulomach: Tests seems to indicate we're not.

kubernetes/lib/charms/mysql/v0/mysql.py
sinclert-canonical: Yep, same as the `client` intermediate variable. Can revert if you think this is none-sense.

kubernetes/lib/charms/mysql/v0/mysql.py
sinclert-canonical: I think you are pointing in the right direction.

According to the 1st _important_ note in this MySQL [page](https://dev.mysql.com/doc/refman/8.4/en/group-replication-configuring-ssl-for-recovery.html), when the `mysql` communication stack is used, the `group_replication_ssl_mode` is the one that triggers the ON/OFF behaviour depending on its value. The `group_replication_recovery_use_ssl` variable seems to be used only for the recovery procedures.

kubernetes/tests/integration/integration/test_tls.py
sinclert-canonical: Let me iterate a bit on it.

kubernetes/lib/charms/mysql/v0/mysql.py
paulomach: no big deal

paulomach (approved):
Good stuff, thank you

kubernetes/tests/integration/integration/test_tls.py
sinclert-canonical: Added delays between juju status checks in commit https://github.com/canonical/mysql-operators/pull/271/commits/0c3f80cfd308a25564a4192b27eb60a7b60ed260.

astrojuanlu (approved):

kubernetes/tests/integration/integration/test_tls.py
astrojuanlu: +1

astrojuanlu (approved):
Did a second pass - still LGTM, just echoed @paulomach 's review comment

kubernetes/tests/integration/integration/test_tls.py
sinclert-canonical: This PR was tweaked a couple of time since this comment was left, so it _may_ not be applicable now. In a nutshell, the `wait_for_apps_status` helper allow us to wait until both:

1. The applications are in the desired status (i.e. `active`).
2. The unit agents are done executing (i.e. `idle`).

The second point is quite relevant for these tests, as during the rolling-ops operations there will always be a unit whose agent is _executing_. Having 3 consecutive checks, 5 seconds apart from each other, where all unit agents are _idle_, is more than enough to make sure the group-replication encryption had enough time to re-configure.

sinclert-canonical:
🗞️ PR branch rebased from `8.4/edge` branch.

PR #291, opened 2026-05-04
[MISC] 8.4 - Consolidate experimental configs

sinclert-canonical:
This PR consolidates the `max-connections` config option after our conversation with Managed Solutions. This setting is heavily used by SunBeam folks, and we should make it official.

### DIFF
--- a/kubernetes/config.yaml
+++ b/kubernetes/config.yaml
@@ -10,6 +10,12 @@ options:
       Optional - Name for async replication cluster set, set once at deployment.
       On `recreate-clster` action call, the cluster set name will be re-generated automatically.
     type: string
+  max-connections:
+    type: int
+    description: |
+      Maximum number of connections allowed to the MySQL server.
+      When set max-connections value take precedence over the memory utilization
+      against innodb_buffer_pool_size.
   profile:
     description: |
       Profile representing the scope of deployment, and used to be able to enable high-level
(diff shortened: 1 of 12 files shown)


astrojuanlu (approved):

astrojuanlu:
Are we still pursuing this? cc @paulomach

sinclert-canonical:
The status of this PR needs to be resolved before promoting the first MySQL 8.4 stable revision. Do we want to make the config permanent? Do we rather remove it altogether?

@paulomach

paulomach:
> The status of this PR needs to be resolved before promoting the first MySQL 8.4 stable revision. Do we want to make the config permanent? Do we rather remove it altogether?
> 
> @paulomach

To not leave publicly unanswered, we are checking alternatives setup with mysql-router before doing an informed decision.

paulomach:
> > The status of this PR needs to be resolved before promoting the first MySQL 8.4 stable revision. Do we want to make the config permanent? Do we rather remove it altogether?
> > @paulomach
> 
> To not leave publicly unanswered, we are checking alternatives setup with mysql-router before doing an informed decision.

mysql-router does not seems to work as we expected for connection sharing. Validating with upstream folks.

sinclert-canonical:
🗞️ PR branch rebased from `8.4/edge` branch. cc @paulomach

PR #301, opened 2026-05-13
[MISC] Tailored AI bugfix agent and workshop file

paulomach:
## Issue

High confidence bug fixing can be queued for too long.

## Solution

Create tailored agent that will produce fix PR for bugs when there's a high confidence that it can be fixed by the agent.
The prompt:

Create an agentic workflow as a github action for this repository that scans open issues with the "bug" label and propose PRs where the solution confidence is high. The workflow should
- able to be manually triggered
- run on a weekly schedule.
- Use claude opus 4.5 model for the agentic call.
- Skip the analysis if a PR is already created for the issue.
- base branch needs to be inferred from the bug report, either per revision correlation as tagged in the repository (e.g. revision 408 is tagged as mysql-k8s/rev408 in 8.4/edge branch) or explicitly mentioned in the issue (e.g. "mysql 8.0 fails to bootstrap" relates to 8.0/edge branch)

### DIFF
--- a/workshop.yaml
+++ b/workshop.yaml
@@ -1,0 +1,8 @@
+name: hackathon
+base: ubuntu@24.04
+sdks:
+  - name: opencode
+    channel: latest/edge
+  - name: uv
+  - name: claude-code
+    channel: latest/edge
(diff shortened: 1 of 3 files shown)


workshop.yaml
```diff
@@ -0,0 +1,8 @@
+name: hackathon
```
astrojuanlu, line 1: We could extract the `workshop.yaml` to its own file, I'd like to start using it more

paulomach: I honestly like it - though I wonder if it's also a place to keep personal preferences

paulomach:
Closing this one as it was part of a testing/hackathon

PR #272, opened 2026-04-27
[DPE-9963] Fix failed K8s initialization

astrojuanlu:
## Issue
- Fixes the resetting of the MySQL directories
- Adds more observability to some critical parts of the code

Fix gh-267

## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.

PR #304, opened 2026-05-20
[MISC] (8.0/edge) Improve logging of get-cluster-status

astrojuanlu:
## Issue

This is a critical operation needed for troubleshooting that had obscure failure modes. In this PR I deleted some dead code and improved the logging so that, on DEBUG level, we can know immediately what went wrong.

Example logs:

```
unit-mysql-k8s-0: 21:04:49.208 DEBUG unit.mysql-k8s/0.juju-log Failed when fetching cluster status: Cannot set LC_ALL to locale en_US.UTF-8: No such file or directory
MySQL Error 2003 (HY000): Can't connect to MySQL server on 'mysql-k8s-0.mysql-k8s-endpoints.testing.svc.cluster.local.:33062' (111)                                                                                  
...
unit-mysql-k8s-0: 21:04:59.578 DEBUG unit.mysql-k8s/0.juju-log Failed when fetching cluster status: Shell Error (51314): Dba.get_cluster: This function is not available through a session to an InnoDB Cluster that belongs to an InnoDB ClusterSet but is not ONLINE
```

## Checklist

### DIFF
--- a/kubernetes/lib/charms/mysql/v0/mysql.py
+++ b/kubernetes/lib/charms/mysql/v0/mysql.py
@@ -1960,6 +1960,8 @@ class MySQLBase(ABC):
         """Get the cluster status dictionary."""
         if not from_instance:
             from_instance = self.instance_address
+        if extended is None:
+            extended = False
 
         client = MySQLClusterClient(
             executor=self._build_instance_tcp_executor(from_instance),
--- a/kubernetes/lib/charms/mysql/v0/mysql.py
+++ b/kubernetes/lib/charms/mysql/v0/mysql.py
@@ -1950,11 +1950,6 @@ class MySQLBase(ABC):
         else:
             return unit_label in labels
 
-    @retry(
-        wait=wait_fixed(2),
-        stop=stop_after_attempt(3),
-        retry=retry_if_exception_type(ExecutionError),
-    )
     def get_cluster_status(
         self, from_instance: str | None = None, extended: bool | None = False
     ) -> dict | None:
--- a/kubernetes/lib/charms/mysql/v0/mysql.py
+++ b/kubernetes/lib/charms/mysql/v0/mysql.py
@@ -78,6 +78,24 @@ from typing import (
 
 import ops
 from charms.data_platform_libs.v0.data_interfaces import DataPeerData, DataPeerUnitData
+from constants import (
+    BACKUPS_PASSWORD_KEY,
+    BACKUPS_USERNAME,
+    CHARMED_MYSQL_PITR_HELPER,
+    CLUSTER_ADMIN_PASSWORD_KEY,
+    CLUSTER_ADMIN_USERNAME,
+    COS_AGENT_RELATION_NAME,
+    GR_MAX_MEMBERS,
+    MONITORING_PASSWORD_KEY,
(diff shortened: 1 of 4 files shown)


kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -1949,25 +1950,23 @@ def instance_belongs_to_cluster(self, unit_label: str) -> bool:
         else:
             return unit_label in labels
 
-    @retry(
-        wait=wait_fixed(2),
-        stop=stop_after_attempt(3),
-        retry=retry_if_exception_type(ExecutionError),
-    )
```
astrojuanlu, line 1956: Notice that this method was never raising. Therefore, this has been dead code for some time.

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -1949,25 +1950,23 @@ def instance_belongs_to_cluster(self, unit_label: str) -> bool:
...
     def get_cluster_status(
         self, from_instance: str | None = None, extended: bool | None = False
     ) -> dict | None:
         """Get the cluster status dictionary."""
         if not from_instance:
             from_instance = self.instance_address
+        if extended is None:
+            extended = False
```
astrojuanlu: This is a typing issue but potentially a logic one too. `fetch_cluster_status` expects a `bool`, whereas this parameter is `bool | None`.

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -1949,25 +1950,23 @@ def instance_belongs_to_cluster(self, unit_label: str) -> bool:
...
 
         client = MySQLClusterClient(
             executor=self._build_instance_tcp_executor(from_instance),
         )
 
         try:
             status = client.fetch_cluster_status(self.cluster_name, extended)
-        except ExecutionError:
+        except ExecutionError as exc:
+            logger.debug(f"Failed when fetching cluster status: {exc}")
```
astrojuanlu, line 1966: This could have been in mysql-shell-client, but @sinclert-canonical expressed his desire to avoid driving shell client decisions by how it's used downstream, and I agree.

kubernetes/src/mysql_k8s_executor.py
```diff
@@ -45,13 +45,16 @@ def _connection_args(self) -> list[str]:
                 f"--user={self._conn_details.username}",
             ]
 
-    def _parse_error(self, output: str) -> dict:
```
astrojuanlu, line 48: This ports the changes done in 8.4/edge in #272.

kubernetes/lib/charms/mysql/v0/mysql.py
sinclert-canonical: Good catch!

I think the function signature is all wrong, `bool` values do not need to default to `None` if not provided, as that is a _falsy_ value which could be simplified by setting it to `False` directly

sinclert-canonical (commented):
Please do the exact same changes in the VM library.

paulomach (approved):
Good idea

sinclert-canonical (approved):

PR #312, opened 2026-05-21
[DPE-10193] 8.0 - Fix Terraform expose block

sinclert-canonical:
This PR fixes the hard-coding of the `expose` block done in PR https://github.com/canonical/mysql-operators/pull/200, by making it optional.

The current Terraform definition was flawed, as exposing an endpoint / space / CIDR trigger different actions depending on the substrate: for VM it just requires to open certain ports, but for Kubernetes it requires the creation of an Ingress resource, which needs to be configured by passing the `juju-external-hostname` config.

### Additional changes
- Added missing LICENSE headers to all the Terraform files.
- Limited Juju provider version to be `1.X.Y`, now that version `2.0.0` is in release candidate phase.

PR #313, opened 2026-05-21
[DPE-10193] 8.4 - Fix Terraform expose block

sinclert-canonical:
This PR fixes the hard-coding of the `expose` block done in PR https://github.com/canonical/mysql-operators/pull/201, by making it optional.

The current Terraform definition was flawed, as exposing an endpoint / space / CIDR trigger different actions depending on the substrate: for VM it just requires to open certain ports, but for Kubernetes it requires the creation of an Ingress resource, which needs to be configured by passing the `juju-external-hostname` config.

### Additional changes
- Added missing LICENSE headers to all the Terraform files.
- Limited Juju provider version to be `1.X.Y`, now that version `2.0.0` is in release candidate phase.

### DIFF
--- a/kubernetes/terraform/main.tf
+++ b/kubernetes/terraform/main.tf
@@ -27,7 +27,7 @@ resource "juju_application" "mysql_server" {
     content {
       endpoints = lookup(expose.value, "endpoints", null)
       spaces    = lookup(expose.value, "spaces", null)
-      cidrs     = lookup(expose.value, "cirds", null)
+      cidrs     = lookup(expose.value, "cidrs", null)
     }
   }
 }
--- a/kubernetes/terraform/main.tf
+++ b/kubernetes/terraform/main.tf
@@ -1,3 +1,6 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
 resource "juju_application" "mysql_server" {
   model_uuid = var.model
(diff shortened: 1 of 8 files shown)


kubernetes/terraform/main.tf
```diff
@@ -18,7 +21,13 @@ resource "juju_application" "mysql_server" {
...
 
-  expose {
-    endpoints = "database"
+  dynamic "expose" {
+    for_each = var.expose != null ? [var.expose] : []
+
+    content {
+      endpoints = lookup(expose.value, "endpoints", null)
+      spaces    = lookup(expose.value, "spaces", null)
+      cidrs     = lookup(expose.value, "cirds", null)
```
paulomach:
```suggestion
      cidrs     = lookup(expose.value, "cidrs", null)
```

sinclert-canonical: Good catch! Let's avoid another `gig` situation 😆 

Fixed in commit https://github.com/canonical/mysql-operators/pull/313/commits/a3de3fa5d3b7dea6c15eafb4a0359d9f3b5edfbc.

paulomach (approved):
Ok, [but](https://github.com/canonical/mysql-operators/pull/312#pullrequestreview-4348288593)

astrojuanlu (approved):

PR #315, opened 2026-05-22
[MISC] 8.4 - Centralize relation-data test helper

sinclert-canonical:
This PR proposes a simplification regarding how the integration tests get relation data. This logic is duplicated across:

- Kubernetes:
  - `helpers.py` module (see [code](https://github.com/canonical/mysql-operators/blob/d9cfafa737a9502ef49417b51cf72d82f4d24fd4/kubernetes/tests/integration/helpers_ha.py#L327-L351)).
  - `test_tls.py` module (see [code](https://github.com/canonical/mysql-operators/blob/d9cfafa737a9502ef49417b51cf72d82f4d24fd4/kubernetes/tests/integration/integration/test_tls.py#L163-L183)).
- Machines:
  - `test_database.py` module (see [code](https://github.com/canonical/mysql-operators/blob/d9cfafa737a9502ef49417b51cf72d82f4d24fd4/machines/tests/integration/integration/relations/test_database.py#L269-L293)).
  - `test_tls.py` module (see [code](https://github.com/canonical/mysql-operators/blob/d9cfafa737a9502ef49417b51cf72d82f4d24fd4/machines/tests/integration/integration/test_tls.py#L162-L182)).

### Additional changes

- Fixed MySQL Test app status refactor leftover in the `test_spaced_db.py` module.
- Checked whether the CA files are present, for both TLS certificates relations.

PR #318, opened 2026-05-22
[DPE-7899] 8.4 - Migrate to TLS v4 (III)

sinclert-canonical:
This PR exposes config options in order to rotate the private key that gets passed on TLS certificate generation.

### Approach

The overall approach has been ported from the Charmed ETCD operator, where the desired private-keys need to be defined within a Juju secrets, to later be passed down to the application configuration via secret URI. The private keys are validated and passed down to the TLS relation handlers.

### Additional changes
- Bumped `charmlibs-rollingops` to version `1.1.0`, allowing us to get rid of a work-around the [previous PR](https://github.com/canonical/mysql-operators/pull/271).

---

Blocked by https://github.com/canonical/mysql-operators/pull/316.

Depends on https://github.com/canonical/mysql-operators/pull/315.

### DIFF
--- a/docs/how-to/tls/manage-private-keys.md
+++ b/docs/how-to/tls/manage-private-keys.md
@@ -1,0 +1,59 @@
+---
+myst:
+  html_meta:
+    description: "Learn how to manage TLS private keys for Charmed MySQL using the self-signed-certificates operator."
+---
+
+# How to manage private keys
+
+You can manage private keys used by the charm to generate the certificate signing requests (CSR) by storing the private key in a [juju secret](https://canonical-juju.readthedocs-hosted.com/en/latest/user/reference/secret/) and then referencing the secret in the [charm configuration](https://canonical-juju.readthedocs-hosted.com/en/latest/user/howto/manage-applications/#configure-an-application). The keys need to be PEM formatted, base64 encoded and at least 2048 bits long.
+
+## Store the private key in a Juju secret
+
+To store the private key in a juju secret, run the following command:
+
+```shell
+juju add-secret tls-client-private-key private-key=$(base64 -w0 private-key.key)
+juju add-secret tls-peer-private-key private-key=$(base64 -w0 private-key.key)
+```
+
+## Reference the secret in the charm configuration
+
+You can use the secret ID from the output to reference the secret in the charm configuration.
+Now that the secret is stored, you can grant the secret to the application using the following commands:
+
+````{tab-set}
+```{tab-item} VM
+:sync: vm
+
+    juju grant-secret tls-client-private-key mysql
+    juju config mysql tls-client-private-key=<SECRET_ID>
+```
+
+```{tab-item} K8s
+:sync: k8s
+
+    juju grant-secret tls-client-private-key mysql-k8s
+    juju config mysql-k8s tls-client-private-key=<SECRET_ID>
+```
+````
+
+Setting the private key for the peer-to-peer communication is similar to the client-to-server communication:
+
+````{tab-set}
+```{tab-item} VM
+:sync: vm
+
+    juju grant-secret tls-peer-private-key mysql
+    juju config mysql tls-peer-private-key=<SECRET_ID>
+```
+
+```{tab-item} K8s
+:sync: k8s
+
+    juju grant-secret tls-peer-private-key mysql-k8s
+    juju config mysql-k8s tls-peer-private-key=<SECRET_ID>
+```
+````
+
+Once the configuration is set, the charm will use the private key stored in the secret to generate new certificate signing requests (CSR) to acquire new certificates from the TLS provider.
--- a/kubernetes/pyproject.toml
+++ b/kubernetes/pyproject.toml
@@ -6,7 +6,7 @@ main = [
     "boto3~=1.28",
     "charm-refresh~=3.1.1",
     "charmlibs-interfaces-tls-certificates~=1.0",
-    "charmlibs-rollingops~=1.0",
+    "charmlibs-rollingops~=1.1",
     "jinja2~=3.1",
     "lightkube~=0.19.0",
     "object-storage-charmlib~=1.0.0",
@@ -22,10 +22,6 @@ charm-libs = [
     # mysql/v0/*.py"
     "charm-refresh~=3.1.1",
     "mysql_shell_client[contrib]~=1.0",
-    # tls_certificates_interface/v1/tls_certificates.py
-    # tls_certificates lib uses a feature only available in cryptography >=42.0.5
-    "cryptography>=42.0.5",
-    "jsonschema",
     # loki_k8s/v0/loki_push_api.py and prometheus_k8s/v0/prometheus_scrape.py
     "cosl>=0.0.50",
 ]
@@ -56,6 +52,7 @@ integration = [
     "kubernetes~=27.2.0",
     "allure-pytest~=2.13",
     "allure-pytest-default-results~=0.1.2",
+    "cryptography~=47.0",
     "jubilant~=1.4.0",
     "tomli~=2.3",
     "tomli-w~=1.2",
--- a/kubernetes/src/charm.py
+++ b/kubernetes/src/charm.py
@@ -10,7 +10,6 @@ from ops.main import main
 if is_wrong_architecture() and __name__ == "__main__":
     main(WrongArchitectureWarningCharm)
 
-import json
 import logging
 import random
 from time import sleep
@@ -565,8 +564,10 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
             logger.info("Skipping group replication restart")
             return OperationResult.RETRY_RELEASE
 
-        ongoing_ops = [relation.data[unit].get("operations", "[]") for unit in self.peers.units]
-        ongoing_ops = [json.loads(op) for op in ongoing_ops]
+        ongoing_ops = [
+            self.rolling_ops.is_waiting_callback("replication", unit.name)
+            for unit in self.peers.units
+        ]
 
         if self.is_unit_primary and any(ongoing_ops):
             logger.info("Skipping group replication restart")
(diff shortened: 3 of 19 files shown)


docs/how-to/tls/manage-private-keys.md
```diff
@@ -0,0 +1,59 @@
+---
+myst:
+  html_meta:
+    description: "Learn how to manage TLS private keys for Charmed MySQL using the self-signed-certificates operator."
+---
+
+# How to manage private keys
```
paulomach, line 7: Add note on the private key format?

paulomach: https://github.com/canonical/charmlibs/blob/0e15dcee31f73ded4b15aea8874c26a46990daf7/interfaces/tls-certificates/src/charmlibs/interfaces/tls_certificates/_tls_certificates.py#L391

sinclert-canonical: Done in commit https://github.com/canonical/mysql-operators/pull/318/commits/64c17e311036815084bf488cb4595b1647c54fa2 + rebased from `8.4/edge`.

kubernetes/src/charm.py
```diff
@@ -565,8 +564,10 @@ def _restart_group_replication(self) -> OperationResult:
             logger.info("Skipping group replication restart")
             return OperationResult.RETRY_RELEASE
 
-        ongoing_ops = [relation.data[unit].get("operations", "[]") for unit in self.peers.units]
-        ongoing_ops = [json.loads(op) for op in ongoing_ops]
+        ongoing_ops = [
+            self.rolling_ops.is_waiting_callback("replication", unit.name)
+            for unit in self.peers.units
+        ]
```
sinclert-canonical, line 570: This is **highly needed**, as iterating through the list of operations and checking whether there is _something_ is not technically correct: multiple operations could be enqueued, we need to filter by the specific callback we care about (in our case: `replication`).

machines/src/charm.py
```diff
@@ -1049,8 +1054,10 @@ def _restart_group_replication(self) -> OperationResult:
             logger.info("Skipping group replication restart")
             return OperationResult.RETRY_RELEASE
 
-        ongoing_ops = [relation.data[unit].get("operations", "[]") for unit in self.peers.units]
-        ongoing_ops = [json.loads(op) for op in ongoing_ops]
+        ongoing_ops = [
+            self.rolling_ops.is_waiting_callback("replication", unit.name)
+            for unit in self.peers.units
+        ]
```
sinclert-canonical, line 1057: This is **highly needed**, as iterating through the list of operations and checking whether there is _something_ is not technically correct: multiple operations could be enqueued, we need to filter by the specific callback we care about (in our case: `replication`).

kubernetes/pyproject.toml
```diff
@@ -22,10 +22,6 @@ charm-libs = [
     # mysql/v0/*.py"
     "charm-refresh~=3.1.1",
     "mysql_shell_client~=0.7",
-    # tls_certificates_interface/v1/tls_certificates.py
-    # tls_certificates lib uses a feature only available in cryptography >=42.0.5
-    "cryptography>=42.0.5",
-    "jsonschema",
```
sinclert-canonical, line 28: We forgot to remove these explicit dependencies when migrated away from the TLS charmlibs.

machines/pyproject.toml
```diff
@@ -26,9 +26,6 @@ charm-libs = [
     # mysql/v0/*.py"
     "charm-refresh~=3.1.1",
     "mysql_shell_client~=0.7",
-    # tls_certificates_interface/v2/tls_certificates.py
-    "cryptography>=42.0.5",
-    "jsonschema",
```
sinclert-canonical, line 31: We forgot to remove these explicit dependencies when migrated away from the TLS charmlibs.

machines/src/relations/tls.py
```diff
@@ -171,6 +177,47 @@ def _get_peer_tls_files(self) -> tuple[str | None, str | None, str | None]:
...
+            logger.error(f"Secret {secret_id} does not contain a private key.")
+            return None
+
+        if re.match(r"(-+(BEGIN|END) [A-Z ]+-+)", private_key):
+            logger.error(f"Secret {secret_id} must be base64 encoded.")
+            return None
+
+        try:
+            private_key = base64.b64decode(private_key).decode("utf-8").strip()
+        except UnicodeDecodeError as e:
```
paulomach: the decode can raise `binascii.Error` also if the private_key is malformed, we should catch it
```suggestion
        except (UnicodeDecodeError, binascii.Error) as e:
```

machines/src/charm.py
```diff
@@ -298,6 +297,12 @@ def _on_config_changed(self, _) -> None:
         # Override log rotation
         self.log_rotation_setup.setup()
 
+        # Rotate TLS keys
+        if self.config.tls_client_private_key:
+            self.tls.client_certificates_refresh_event.emit()
+        if self.config.tls_peer_private_key:
+            self.tls.peer_certificates_refresh_event.emit()
```
paulomach: We are not controlling for other config changes, i.e. every config change will trigger rotation even if keys have not changed.
One way is to track secret id+revision (first [revision is 1](https://github.com/juju/juju/blob/9adf9855a8a0d70e6b54c2012c33beb33c803852/domain/secret/state/state.go#L673-L697)) if has not changed, skip the event emit

kubernetes/src/charm.py
```diff
@@ -694,6 +695,12 @@ def _on_config_changed(self, _: EventBase) -> None:
         # Override log rotation
         self.log_rotate_setup.setup()
 
+        # Rotate TLS keys
+        if self.config.tls_client_private_key:
+            self.tls.client_certificates_refresh_event.emit()
+        if self.config.tls_peer_private_key:
+            self.tls.peer_certificates_refresh_event.emit()
+
```
paulomach: same as https://github.com/canonical/mysql-operators/pull/318/changes#r3312858268

paulomach (changes requested):
Great work. We just need a way to avoid unnecessary rotations

machines/src/charm.py
sinclert-canonical: Great catch!

This is something I saw in the Charmed ETCD implementation, but disregarded as _too complex_ (they deal with an entirely state machine in the application data-bag). I think we can implement _some_ of the changes they have: at least some kind of unit data-bag state preservation, so we know when a new private key is given, and when it is the same one as before.

WDYT?

machines/src/relations/tls.py
sinclert-canonical: Done in commits https://github.com/canonical/mysql-operators/pull/318/commits/937d23bdb512185fefc747680b8afca290c07f45 + https://github.com/canonical/mysql-operators/pull/318/commits/3425be3d6d6fb76cfa965904d35d334dfd415f76.

machines/src/charm.py
sinclert-canonical: Done in commits https://github.com/canonical/mysql-operators/pull/318/commits/937d23bdb512185fefc747680b8afca290c07f45 + https://github.com/canonical/mysql-operators/pull/318/commits/3425be3d6d6fb76cfa965904d35d334dfd415f76.

paulomach (approved):
Great! Thanks for addressing the comments

kubernetes/src/relations/tls.py
```diff
@@ -258,6 +306,9 @@ def _on_client_relation_broken(self, _: EventBase) -> None:
             self.charm.unit.status = BlockedStatus("Failed to disable client TLS.")
             return
 
+        if self.charm.unit.is_leader():
+            del self.charm.app_peer_data["client-private-key"]
```
astrojuanlu, line 310: What would be the effect of this not happening? For example if a node goes down and there's not enough time for the charm to handle the client-relation-broken event. Should we re-trigger the event until this is done, or is it not that critical?

kubernetes/src/charm.py
astrojuanlu: Not a new issue introduced by this code, but I notice that we regularly assume `self.peers` will be not `None`. Is there any situation in which this wouldn't be true?

kubernetes/tests/integration/integration/test_tls_private_key.py
```diff
@@ -0,0 +1,184 @@
+# Copyright 2026 Canonical Ltd.
```
astrojuanlu, line 1: Is it worth adding a test that disabling TLS works well?

astrojuanlu (approved):
Approved with some questions, thank you!

kubernetes/src/relations/tls.py
sinclert-canonical: This cleanup is done just to prevent data-bag stale values. In an scenario where:
- The TLS relation was disabled.
- The App still have some units around.
- The App data-bag key does not get cleanup.

Both newly spawned units + existing ones will have TLS functionality disabled. Only when the same type of TLS relation gets re-related (either _client_ or _peer_ one), the units will look for config private key (see [K8s code](https://github.com/canonical/mysql-operators/blob/3425be3d6d6fb76cfa965904d35d334dfd415f76/kubernetes/src/relations/tls.py#L81) + [VM code](https://github.com/canonical/mysql-operators/blob/3425be3d6d6fb76cfa965904d35d334dfd415f76/machines/src/relations/tls.py#L79)) and start using it immediately.

---

In short: the previously configured private key will get picked up by all units, **upon TLS re-relation**.

kubernetes/tests/integration/integration/test_tls_private_key.py
sinclert-canonical: Could you elaborate? Just disabling, or disabling + enabling?

astrojuanlu: Disabling and making sure that connections still work. I think the disabling + enabling might be too much.

PR #322, opened 2026-05-26
[MISC] (8.4) Fix condition for publishing Allure results (back to 24.04)

astrojuanlu:
## Issue

Partially reverts #213.

Should have included this in #306.

## Checklist

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -152,7 +152,7 @@ jobs:
         # Only upload results from one spread system & one spread variant
         # Allure can only process one result per pytest test ID. If parameterization is done via
         # spread instead of pytest, there will be overlapping pytest test IDs.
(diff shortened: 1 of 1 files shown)


paulomach (approved):

sinclert-canonical (approved):
I wonder whether we should remove that condition altogether... fine for now.

astrojuanlu:
> I wonder whether we should remove that condition altogether... fine for now.

Explanation from @carlcsaposs-canonical from the last time we thought about this:

> allure needs unique test case IDs, otherwise it thinks test results are multiple retries of the same tests & it won't display the data as you'd expect
> because parameterization is done via spread (systems and variants), we have overlapping test case IDs unless we limit to a single system and variant
> would recommend replacing 24.04 with 22.04 instead of trimming or allure will break with arm results

https://github.com/canonical/mysql-operators/blob/00cbad43ad877872e24db0cbb9e389193116f2bb/.github/workflows/integration_test.yaml#L152-L154

See also `https://github.com/canonical/mysql-operators/pull/213/changes/116f3e4e9af94b99ea0701e7505a35fa55818ec1#r3044253684`

PR #332, opened 2026-05-28
[MISC] (8.0/edge) Fix full cluster recovery when async related

paulomach:
## Issue

Found this issue while testing host reboot recovery and had an async setup in the k8s cluster.
Full cluster recovery is triggered only when the operator detects all units in a cluster are `OFFLINE`.
When in async replication, the leader unit evaluates the async state to prevent doing recovery while the module is mid setup.
Given the unit is offline (no access to the cluster-set), the evaluation fails, preventing the instance to execute the cluster recovery (it's the leader after all).

## Solution

- Only evaluate the async state if the unit is not offline, otherwise, trigger the recovery as usual.
- Allow for async state to be `idle` also when async state is `OK_NOT_CONSISTENT`

## Testing

I am not adding test for this as it will be fairly complex and not a common use case just yet.
To test it manually, one can deploy two 3 node cluster, async relate them and put one or both cluster offline (by killing all pods or reboot all hosts).
Operation should resume


machines/lib/charms/mysql/v0/async_replication.py
```diff
@@ -57,7 +57,7 @@
 # The unique Charmhub library identifier, never change it
 LIBID = "4de21f1a022c4e2c87ac8e672ec16f6a"
 LIBAPI = 0
-LIBPATCH = 13
+LIBPATCH = 12
```
paulomach: cp from k8s to machines

sinclert-canonical (approved):
This should be easily portable to the `8.4/edge` branch, given the reduced patch size.

astrojuanlu (approved):

machines/lib/charms/mysql/v0/async_replication.py
```diff
@@ -57,7 +57,7 @@
...
```
sinclert-canonical, line 61: I assume this is a typo after rebase. Why would we reduce the charmlib patch version? 🤔

paulomach: e6f53ca94

PR #335, opened 2026-06-02
[MISC] fix user deletion on legacy

paulomach:
Removing the last `mysql-root` legacy relation never cleaned up its
users and left the unit stuck. Two bugs combined:

- `delete_users_with_label` built `SELECT user.user, user.host` +
  `FROM mysql.user ...` with no space between the implicitly
  concatenated literals, producing `user.hostFROM` -- the query always
  failed. The unit test had encoded the same typo, so it never caught
  it.
- `_on_mysql_root_relation_broken` caught `MySQLDeleteUsersForUnitError`
  while `delete_users_with_label` raises `MySQLDeleteUsersWithLabelError`,
  so the (now reachable) error propagated and put the hook into an error
  loop instead of setting BlockedStatus.

Fix the SQL spacing, catch the correct exception, and update the test.

Co-Authored-By: Claude Opus 4.8

### DIFF
--- a/kubernetes/src/mysql_k8s_helpers.py
+++ b/kubernetes/src/mysql_k8s_helpers.py
@@ -569,7 +569,7 @@ class MySQL(MySQLBase):
 
         get_label_users_selector = f'%"{label_name}": "{label_value}"%'
         get_label_users_query = (
-            "SELECT user.user, user.host"
+            "SELECT user.user, user.host "  # noqa: S608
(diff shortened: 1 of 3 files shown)


sinclert-canonical (approved):
> The query always failed. The unit test had encoded the same typo, so it never caught it.

My bad 🤦🏻‍♂️

astrojuanlu (approved):

PR #337, opened 2026-06-02
[DPE-7872] 8.4 - Move service classes to own folder

sinclert-canonical:
This PR proposes a refactorization of the background services related classes in order to distinguish them from the charm main observers / handlers. This work is recommended before introducing any new kind of service (i.e. topology observer).

The structure would be:
```log
src/
    ...
    services/
        events.py   <-- Where custom events are grouped together
        managers/   <-- Where background services managers are stored.
            ...
        observers/  <-- Where background services observes are stored.
            ...
    ...
```

### Additional changes:
- Ignored 3 specific linting rules in `__init__.py` files. Let me know what you think, it is easily revertible.


paulomach (approved):

kubernetes/src/services/managers/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+from .log_rotate_manager import LogRotateManager
```
astrojuanlu, line 4: Can we add

```
__all__ = ["LogRotateManager"]
```

and similar to these `__init__.py` files instead of ignoring F401?

astrojuanlu (approved):

kubernetes/src/services/managers/__init__.py
sinclert-canonical: I am not that familiar with the use of `__all__` attribute, and I am worried a string identifier may be a source of bugs, if a typo gets introduced. WDYT about doing what the `ruff` linter recommends?

```diff
- from .log_rotate_manager import LogRotateManager
+ from .log_rotate_manager import LogRotateManager as LogRotateManager
```

astrojuanlu: Luckily, despite being a list of strings, linters catch errors there:

image

image

image

Further info https://docs.python.org/3/reference/simple_stmts.html#module.__all__

Not the biggest fan of `import X as X` but at least it makes the linter happy, so I'd begrudgingly accept 😄

sinclert-canonical: Done in commit https://github.com/canonical/mysql-operators/pull/337/commits/c98e09af60070f3f06cc9735418b0a230efa6b3e.

PR #338, opened 2026-06-02
[DPE-9259] Safe auto-recover from no_quorum

### LINKED ISSUE #21: [k8s] The MySQL cluster is not recovered automatically under certain scenarios.
Killing multiple MySQL pods at the same time without waiting for them to fully come online will put the cluster in an unhealthy state, but it will not trigger the reboot cluster from complete outage flow to recover the cluster.
This will also happen during a networking outage, which in this case was simulated by taking down all NICs on all the microk8s machines in a single AZ.

## Steps to reproduce
1. By killing pods.
  a. Kill multiple mysql pods at the same time.
    ```bash
    kubectl delete mysql-0
    kubectl delete mysql-1
    ```
2. By simulating a network outage.
  a. Take down all NICs of the microk8s machines.
    ```bash
    juju machines -m microk8s | awk '/AZ3/ {print $1}' | while read machine; do
      juju ssh $machine '
        for nic in bond0 bondM; do  # adjust as necessary
          sudo ip link set dev $nic down
        done
      '
    done
    ```
  b. Wait 15 minutes for the network outage to affect mysql, then reboot all the machines in the AZ to bring the network back online.

## Expected behavior
The cluster will go offline and the reboot cluster from complete outage flow will be triggered to recover the cluster.

## Actual behavior
The leader unit will not go offline and the reboot cluster flow will not be triggered, leaving the cluster in an inconsistent state.
```
mysql/0          maintenance  idle       10.1.1.10         offline
mysql/1          maintenance  idle       10.1.1.11         offline
mysql/2*         maintenance  idle       10.1.1.12         Unable to get member state
```

## Versions

Operating system: Ubuntu 22.04 Jammy Jellyfish
Juju CLI: 3.4.3
Juju agent: 3.4.3
Charm revision: 153 (channel 8.0/stable)
microk8s: 1.28

## Additional context
To recover the cluster, the `mysqld_safe` Pebble service needs to be restarted inside the leader unit:
```bash
juju ssh --container mysql mysql/leader /charm/bin/pebble restart mysqld_safe
```

paulomach:
Fix #21 

Use the no_quorum->offline->reboot_from_outage path to safe
restore from quorum loss if all nodes become reachable.

TODO:
- ~~test~~


kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -830,8 +830,13 @@ def active_status_message(self) -> str:
...
             return ""
 
-        if self._mysql.is_cluster_replica() is False:
+        is_replica = self._mysql.is_cluster_replica()
+        if is_replica is False:
             return "Primary"
+        if is_replica is None:
+            # cluster-set status unreadable (e.g. lost quorum): a primary that
+            # cannot accept writes. Avoid mislabelling it as "Standby".
+            return "Primary (read-only)"
```
astrojuanlu, line 839: 👍🏼

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2356,6 +2361,13 @@ def start_group_replication(self) -> None:
         with suppress(ExecutionError):
             self._instance_client_tcp.start_instance_replication()
 
+    def is_cluster_in_no_quorum(self) -> bool:
+        """Check whether the cluster has lost quorum as seen from this instance."""
+        status = self.get_cluster_status()
+        if not status:
+            return False
```
astrojuanlu, line 2383: This could be `None`? If `status` is `None`, then maybe it couldn't be retrieved, and so we don't know if cluster is in NO_QUORUM state

kubernetes/src/charm.py
```diff
@@ -844,6 +844,30 @@ def _handle_potential_cluster_crash_scenario(self) -> bool:  # noqa: C901
         self.unit_peer_data["member-state"] = state.lower()
         self.unit_peer_data["member-role"] = role.lower()
 
+        # A surviving member can stay ONLINE in its local view while the
+        # cluster has lost quorum (majority UNREACHABLE). The reboot-from-
+        # complete-outage path below only fires on state == OFFLINE, so
+        # without this the leader stays stuck and the cluster never recovers.
+        if state == InstanceState.ONLINE and self._mysql.is_cluster_in_no_quorum():
+            logger.warning("Cluster has lost quorum.")
```
astrojuanlu:
```suggestion
            logger.warning("Cluster has lost quorum")
```

astrojuanlu: (Periods seem to be inconsistent in more log lines)

kubernetes/lib/charms/mysql/v0/mysql.py
paulomach: It can be `None` which might indicate other issue - i'd rather state `no_quorum`  when we are absolutely sure

sinclert-canonical (approved):
LGTM!

This one is not that easily portable into the `8.4/edge` branch, and we must be careful:
1. The `member-state` unit data-bag key is not being lowercased in 8.4 operators.
2. The `active_status_message` changes need to be put [here](https://github.com/canonical/mysql-operators/blob/5f41a556c5252468c7f720343c0bd72e4385ef8b/kubernetes/lib/charms/mysql/v0/mysql.py#L859-L873) and [here](https://github.com/canonical/mysql-operators/blob/5f41a556c5252468c7f720343c0bd72e4385ef8b/machines/lib/charms/mysql/v0/mysql.py#L858-L872).

astrojuanlu (approved):

astrojuanlu (approved):

astrojuanlu:
Backporting this following @sinclert-canonical 's advice so I can unblock #341 .

astrojuanlu:
#354

kubernetes/src/charm.py
```diff
@@ -844,6 +844,30 @@ def _handle_potential_cluster_crash_scenario(self) -> bool:  # noqa: C901
         self.unit_peer_data["member-state"] = state.lower()
         self.unit_peer_data["member-role"] = role.lower()
 
+        # A surviving member can stay ONLINE in its local view while the
+        # cluster has lost quorum (majority UNREACHABLE). The reboot-from-
+        # complete-outage path below only fires on state == OFFLINE, so
+        # without this the leader stays stuck and the cluster never recovers.
```
sinclert-canonical, line 852: May this be a fix for the MySQL Router 8.2+ _lost of quorum_ test errors we tried to address? AFAIK, the lost-of-quorum safety mechanism got triggered on one unit deployments, but I am not sure whether, even in those cases, the custer status could be `NO_QUORUM`, while the single instance is online 🤔 

Worth testing once these changes get ported and released into the `8.4/edge` channel.

machines/src/charm.py
```diff
@@ -594,6 +594,29 @@ def _on_update_status(self, _) -> None:  # noqa: C901
...
+        if state == InstanceState.ONLINE and self._mysql.is_cluster_in_no_quorum():
+            logger.warning("Cluster has lost quorum")
+            try:
+                # reboot_cluster_from_complete_outage rejects an instance whose
+                # GR is still running; drop it to OFFLINE first.
+                self._mysql.stop_group_replication()
+                if self.unit.is_leader():
+                    # run on leader only for coordinate recovery
+                    logger.warning("Attempting reboot from complete outage")
+                    self._mysql.reboot_from_complete_outage()
```
astrojuanlu, line 619:
```suggestion
                    self._mysql.reboot_from_complete_outage()
                    return False
```

to match the kubernetes charm?

kubernetes/tests/integration/integration/high_availability/test_self_healing_safe_recover_quorum.py
```diff
@@ -0,0 +1,115 @@
...
+    wait_for_apps_status,
+)
+
+MYSQL_APP_NAME = "mysql-k8s"
+MYSQL_TEST_APP_NAME = "mysql-test-app"
+
+MINUTE_SECS = 60
+
+
+def dtest_deploy_highly_available_cluster(juju: Juju, charm: str) -> None:
```
astrojuanlu:
```suggestion
def test_deploy_highly_available_cluster(juju: Juju, charm: str) -> None:
```

machines/src/charm.py
```diff
@@ -594,6 +594,29 @@ def _on_update_status(self, _) -> None:  # noqa: C901
...
+                else:
+                    # set offline state and delegate reboot from outage to leader
+                    logger.warning("Set instance to offline")
+                    self.unit_peer_data["member-state"] = InstanceState.OFFLINE.lower()
+            except MySQLRebootFromCompleteOutageError:
+                logger.error("Failed to reboot cluster from complete outage")
+                self.unit.status = BlockedStatus("failed to recover cluster")
+            return
```
astrojuanlu, line 623:
```suggestion
            return True
```

to match the kubernetes charm?

astrojuanlu (changes requested):

kubernetes/src/charm.py
```diff
@@ -844,6 +844,30 @@ def _handle_potential_cluster_crash_scenario(self) -> bool:  # noqa: C901
...
+                self._mysql.stop_group_replication()
+                if self.unit.is_leader():
+                    # run on leader only for coordinate recovery
+                    logger.warning("Attempting reboot from complete outage")
+                    self._mysql.reboot_from_complete_outage()
+                    return False
+                else:
+                    # set offline state and delegate reboot from outage to leader
+                    logger.warning("Set instance to offline")
+                    self.unit_peer_data["member-state"] = InstanceState.OFFLINE.lower()
```
sinclert-canonical: Apologies for the late feedback, but I do not think this is a good idea.

We are setting a hard-coded member-state that does not come from the MySQL Shell client directly. Even if we want to convey the idea that, from a cluster perspective, this unit is _offline_, we are not using this information inmediately. Better to wait until the next self-healing service run (WIP) or `update-status` event updates the unit data-bag via the standard procedure. It will be set to OFFLINE regardless, given the group-replication stop.

---

In short: do not manually _short-circuit_, let the self-healing loops deal with the unit data-bag.

machines/src/charm.py
```diff
@@ -594,6 +594,29 @@ def _on_update_status(self, _) -> None:  # noqa: C901
...
+                # GR is still running; drop it to OFFLINE first.
+                self._mysql.stop_group_replication()
+                if self.unit.is_leader():
+                    # run on leader only for coordinate recovery
+                    logger.warning("Attempting reboot from complete outage")
+                    self._mysql.reboot_from_complete_outage()
+                else:
+                    # set offline state and delegate reboot from outage to leader
+                    logger.warning("Set instance to offline")
+                    self.unit_peer_data["member-state"] = InstanceState.OFFLINE.lower()
```
sinclert-canonical: Apologies for the late feedback, but I do not think this is a good idea.

We are setting a hard-coded member-state that does not come from the MySQL Shell client directly. Even if we want to convey the idea that, from a cluster perspective, this unit is _offline_, we are not using this information inmediately. Better to wait until the next self-healing service run (WIP) or `update-status` event updates the unit data-bag via the standard procedure. It will be set to OFFLINE regardless, given the group-replication stop.

---

In short: do not manually _short-circuit_, let the self-healing loops deal with the unit data-bag.

sinclert-canonical (dismissed):

machines/src/charm.py
astrojuanlu: When I backported this to 8.4/edge I got confused though https://github.com/canonical/mysql-operators/pull/354#issuecomment-4658010084

sinclert-canonical:
I am going to vacation tomorrow, and won't be available to re-review 📢

The last thing I want is to block this PR progress, so feel free to manually disregard my _request for comment_.

kubernetes/src/charm.py
paulomach: This will happen when 2 units depart simultaneously (no-reelection) and the remaining one was a secondary. So it stays put `ONLINE` but as a replica, with a no quorum cluster.
So is not quorum loss from router perspective, but cluster level quorum loss.

machines/src/charm.py
paulomach: We should return - they are not exactly the same (being more specific, on k8s recovery is delegated to a method and in vm is all in the event handler).
It may not be too hard to reconcile, but that would be out of scope on this PR.

machines/src/charm.py
paulomach: same as https://github.com/canonical/mysql-operators/pull/338#discussion_r3383223104

kubernetes/src/charm.py
paulomach: The short-circuit aims to speed up recovery.
Happy to do it given the speed-up provided by the topology observer

kubernetes/tests/integration/integration/high_availability/test_self_healing_safe_recover_quorum.py
paulomach: efda11c8f

kubernetes/src/charm.py
paulomach: efda11c8f

machines/src/charm.py
paulomach: efda11c8f

kubernetes/tests/integration/integration/high_availability/test_primary_switchover.py
```diff
@@ -116,28 +114,20 @@ def test_cluster_failover_after_majority_loss(juju: Juju) -> None:
...
-            ready=lambda status: all((
-                wait_for_unit_status(app_name, unit_to_promote, "active")(status),
-                wait_for_unit_message(app_name, units_to_kill[0], "OFFLINE")(status),
-                wait_for_unit_message(app_name, units_to_kill[1], "OFFLINE")(status),
-            )),
-            timeout=15 * MINUTE_SECS,
-            delay=15,
+    # ensure no update-status is triggered
+    with update_interval(juju, "30m"):
+        kill_pods(juju, units_to_kill)
```
astrojuanlu: Please have a look at the approach I'm proposing in https://github.com/canonical/mysql-operators/pull/354, I tried the `kill_pods` one but the no_quorum status passed too quickly locally

machines/src/charm.py
astrojuanlu: Fair enough, I proposed a fix in https://github.com/canonical/mysql-operators/pull/354

kubernetes/tests/integration/integration/high_availability/test_primary_switchover.py
paulomach: Done - comments there

astrojuanlu (approved):
Let's go!

marceloneppel (approved):

kubernetes/tests/integration/helpers_ha.py
```diff
@@ -702,3 +702,37 @@ def wait_for_unit_message(app_name: str, unit_name: str, unit_message: str) -> J
...
+
+
+def force_kill_mysqld_service(juju: Juju, unit_name: str) -> None:
+    """SIGKILL mysqld and suppress Pebble's auto-restart.
+
+    Args:
+        juju: The Juju model.
+        unit_name: The name of the unit.
+
+    """
```
sinclert-canonical, line 714: This whole helper seems similar to the logic we have for the process-kill integration test (see [source](https://github.com/canonical/mysql-operators/blob/3af940bf6735fdcec6a208d4974d8cd0fd8f2c06/kubernetes/tests/integration/integration/high_availability/test_self_healing_process_killed.py#L85-L93)), probably even better:

1. It does not leverage `lightkube` to execute commands within the container.
2. It does not have an arbitrary `time.sleep` command.

For the future: it may be worth centralizing this operations in an unified place (as we have done with many other integration test operations), as well as analyze whether we can ditch `lightkube` altogether, and rely on Jubilant to execute commands.

PR #340, opened 2026-06-03
[MISC] Fix node count target

paulomach:
## Issue

On manual rejoin procedure, fallback path of instance_remove->instance_add fails at the remove step.
The node count is done against self, which is offline on the cluster, detecting single node, which in turns set remove path as cluster dissolve (last node in the cluster), which is not the case.

## Solution

When `from_instance` is set, use it for node count.


sinclert-canonical (approved):
Good catch! This should be easily portable to `8.4/edge`.

astrojuanlu (approved):

PR #343, opened 2026-06-04
[DPE-7872] 8.4 - Define self-healing service

sinclert-canonical:
This PR defines a new self-healing service in order to decouple crash-recovery and endpoint updates from the model-level `update-status` event (defaults to 5 min, but it is configurable by the user). The service has the following characteristics:

- Runs a background process in the charm container.
- Emits a custom event every 2 min (120 secs).
- Does **not** perform any validation before emitting.

---

Depends on PR https://github.com/canonical/mysql-operators/pull/337.


paulomach (approved):

astrojuanlu (approved):
Gave this a quick pass after reviewing the companion 8.0/edge PR 👍🏼

Issue #350, opened 2026-06-08
Intermittent "connection details must not be empty" error on mysql-pebble-ready


astrojuanlu:
While working on #341 I found this in the logs:

```
unit-mysql-k8s-1: 10:19:50.298 DEBUG unit.mysql-k8s/1.juju-log root:Emitting Juju event mysql_pebble_ready.                                                                          
unit-mysql-k8s-1: 10:19:50.300 DEBUG unit.mysql-k8s/1.juju-log root:Starting root trace with id='e62dd5c4d5829494fd8f38c69eb8c569'.                                                  
unit-mysql-k8s-1: 10:19:50.372 WARNING unit.mysql-k8s/1.juju-log __main__:Unit DNS domain name is not propagated yet   
unit-mysql-k8s-1: 10:19:50.386 ERROR unit.mysql-k8s/1.juju-log root:Uncaught exception while in charm code:                                                                          
Traceback (most recent call last):                                                                                                                                                   
  File "/var/lib/juju/agents/unit-mysql-k8s-1/charm/src/charm.py", line 1174, in <module>                                                                                            
    main(MySQLOperatorCharm)                                                                                                                                                         
  File "/var/lib/juju/agents/unit-mysql-k8s-1/charm/venv/lib/python3.12/site-packages/ops/main.py", line 39, in main                                                                 
    return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)                                                                                            
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                            
  File "/var/lib/juju/agents/unit-mysql-k8s-1/charm/venv/lib/python3.12/site-packages/ops/_main.py", line 507, in main                                                               
    manager.run()                                                                                                                                                                    
  File "/var/lib/juju/agents/unit-mysql-k8s-1/charm/venv/lib/python3.12/site-packages/ops/_main.py", line 491, in run                           
    self._emit()                                                                                                                                                                     
  File "/var/lib/juju/agents/unit-mysql-k8s-1/charm/venv/lib/python3.12/site-packages/ops/_main.py", line 426, in _emit                                                              
    self._emit_charm_event(self.dispatcher.event_name)                                                                                                                               
[...]

PR #136, opened 2026-03-09
[DPE-8064] Rework app status

sinclert-canonical:
This PR reworks the app status setting by doing a mapping between the MySQL cluster status and the Juju app status:

- When MySQL cluster has > 3 instances and cluster state is `OK_*` -> Active and healthy.
- When MySQL cluster has > 3 instances and cluster state is `OK_NOT_TOLERANCE_*` -> Active but degraded.
- When MySQL cluster is in any other known state -> Blocked.
- When MySQL cluster is in any unknown state -> Maintenance.

#### References:
- MySQL cluster statuses [documentation](https://dev.mysql.com/doc/mysql-shell/8.4/en/monitoring-innodb-cluster.html#:~:text=status%3A%20The%20status%20of%20the%20InnoDB%20Cluster.%20The%20status%20describes%20the%20high%20availability%20provided%20by%20this%20cluster.%20The%20status%20is%20one%20of%20the%20following).

PR #342, opened 2026-06-04
[DPE-7872] 8.0 - Define self-healing service

sinclert-canonical:
This PR defines a new self-healing service in order to decouple crash-recovery and endpoint updates from the model-level `update-status` event (defaults to 5 min, but it is configurable by the user). The service has the following characteristics:

- Runs a background process in the charm container.
- Emits a custom event every 2 min (120 secs).
- Does **not** perform any validation before emitting.

---

Depends on PR https://github.com/canonical/mysql-operators/pull/336.

PR #354, opened 2026-06-09
[DPE-9259] (8.4/edge) Safe auto-recover from no_quorum

astrojuanlu:
## Issue

Use the no_quorum->offline->reboot_from_outage to safe restore from quorum loss when all nodes become reachable

(Backports #338 with #136 in mind)

## Checklist

### DIFF
--- a/kubernetes/lib/charms/mysql/v0/mysql.py
+++ b/kubernetes/lib/charms/mysql/v0/mysql.py
@@ -863,8 +863,13 @@ class MySQLCharmBase(CharmBase, ABC):
         if self.unit_peer_data.get("member-role") != InstanceRole.PRIMARY:
             return ops.ActiveStatus("")
 
-        if self._mysql.is_cluster_replica() is False:
+        is_replica = self._mysql.is_cluster_replica()
+        if is_replica is False:
             return ops.ActiveStatus("Primary")
+        if is_replica is None:
+            # cluster-set status unreadable (e.g. lost quorum): a primary that
+            # cannot accept writes. Avoid mislabelling it as "Standby".
+            return ops.ActiveStatus("Primary (read-only)")
 
         status = self._mysql.get_replica_cluster_status()
         if status == ClusterGlobalStatus.OK:
@@ -2311,6 +2316,13 @@ class MySQLBase(ABC):
         """Start Group replication on the instance."""
         with suppress(ExecutionError):
             self._instance_client_tcp.start_instance_replication()
+
+    def is_cluster_in_no_quorum(self) -> bool:
+        """Check whether the cluster has lost quorum as seen from this instance."""
+        status = self.get_cluster_status()
+        if not status:
+            return False
+        return status["defaultReplicaSet"]["status"] == ClusterStatus.NO_QUORUM
 
     def force_quorum_from_instance(self) -> None:
         """Force quorum from the current instance.
--- a/kubernetes/src/charm.py
+++ b/kubernetes/src/charm.py
@@ -943,8 +943,8 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
 
         Returns:
             bool
-                True if the handling worked correctly and the caller can continue,
-                False otherwise.
+                False if the handling worked correctly and the caller can continue,
+                True otherwise.
 
         """
         single_node_cluster = self.only_one_cluster_node_thats_uninitialized
@@ -1121,7 +1121,9 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
         self.unit_peer_data["member-state"] = state
         self.set_unit_status(self.build_unit_workload_status())
 
-        if self._handle_potential_cluster_crash_scenario(state):
+        # TODO: Logic here is almost the opposite as the machines charm, but not quite
+        # We should review and fix it
+        if not self._handle_potential_cluster_crash_scenario(state):
             self._set_app_status(state)
 
     def _set_app_status(self, state: str) -> None:
--- a/kubernetes/src/charm.py
+++ b/kubernetes/src/charm.py
@@ -938,7 +938,7 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
         self._create_cluster()
         self._mysql.reconcile_binlogs_collection(force_restart=True)
 
-    def _handle_potential_cluster_crash_scenario(self, state: str) -> bool:
+    def _handle_potential_cluster_crash_scenario(self, state: str) -> bool:  # noqa: C901
         """Handle potential full cluster crash scenarios.
 
         Returns:
@@ -946,6 +946,30 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
         """
         single_node_cluster = self.only_one_cluster_node_thats_uninitialized
         if not single_node_cluster and not self.cluster_initialized:
+            return True
+
+        # A surviving member can stay ONLINE in its local view while the
+        # cluster has lost quorum (majority UNREACHABLE). The reboot-from-
+        # complete-outage path below only fires on state == OFFLINE, so
+        # without this the leader stays stuck and the cluster never recovers.
+        if state == InstanceState.ONLINE and self._mysql.is_cluster_in_no_quorum():
+            logger.warning("Cluster has lost quorum")
+            try:
+                # reboot_cluster_from_complete_outage rejects an instance whose
+                # GR is still running; drop it to OFFLINE first.
+                self._mysql.stop_group_replication()
+                if self.unit.is_leader():
+                    # run on leader only for coordinate recovery
+                    logger.warning("Attempting reboot from complete outage")
+                    self._mysql.reboot_from_complete_outage()
+                    return False
+                else:
+                    # set offline state and delegate reboot from outage to leader
+                    logger.warning("Set instance to offline")
+                    self.unit_peer_data["member-state"] = InstanceState.OFFLINE
+            except MySQLRebootFromCompleteOutageError:
+                logger.error("Failed to reboot cluster from complete outage")
+                self.set_unit_status(BlockedStatus("failed to recover cluster"))
             return True
 
         if state == InstanceState.RECOVERING:
--- a/kubernetes/src/charm.py
+++ b/kubernetes/src/charm.py
@@ -963,10 +963,6 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
                     logger.warning("Attempting reboot from complete outage")
                     self._mysql.reboot_from_complete_outage()
                     return False
-                else:
-                    # set offline state and delegate reboot from outage to leader
-                    logger.warning("Set instance to offline")
-                    self.unit_peer_data["member-state"] = InstanceState.OFFLINE
             except MySQLRebootFromCompleteOutageError:
                 logger.error("Failed to reboot cluster from complete outage")
                 self.set_unit_status(BlockedStatus("failed to recover cluster"))
--- a/kubernetes/src/charm.py
+++ b/kubernetes/src/charm.py
@@ -12,6 +12,7 @@ if is_wrong_architecture() and __name__ == "__main__":
 
 import logging
 import random
+import socket
 from time import sleep
 
 import charm_refresh
@@ -400,6 +401,21 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
             raise RuntimeError("Can't get fully qualified domain name for unit")
 
         return dotappend(unit_dns_domain)
+
+    def _all_peers_reachable(self) -> bool:
+        """Return True if all peer units respond on MySQL port 3306.
+
+        Quorum recovery must not run when peers are unreachable — that scenario
+        indicates a network partition where the remote side may still be healthy.
+        """
+        for unit in self.peers.units:
+            address = self.get_unit_address(unit)
+            try:
+                with socket.create_connection((address, 3306), timeout=2):
+                    pass
+            except OSError:
+                return False
+        return True
 
     def is_unit_busy(self) -> bool:
         """Returns whether the unit is busy."""
@@ -956,7 +972,10 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
         # complete-outage path below only fires on state == OFFLINE, so
         # without this the leader stays stuck and the cluster never recovers.
         if state == InstanceState.ONLINE and self._mysql.is_cluster_in_no_quorum():
-            logger.warning("Cluster has lost quorum")
+            logger.warning("Cluster has no quorum")
+            if self.peers.units and not self._all_peers_reachable():
+                logger.warning("Skipping quorum recovery: not all peers reachable")
+                return True
             try:
                 # reboot_cluster_from_complete_outage rejects an instance whose
                 # GR is still running; drop it to OFFLINE first.
--- a/kubernetes/src/charm.py
+++ b/kubernetes/src/charm.py
@@ -942,7 +942,10 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
         """Handle potential full cluster crash scenarios.
 
         Returns:
-            bool indicating whether the caller should return
+            bool
+                True if the handling worked correctly and the caller can continue,
+                False otherwise.
+
         """
         single_node_cluster = self.only_one_cluster_node_thats_uninitialized
         if not single_node_cluster and not self.cluster_initialized:
@@ -1119,9 +1122,7 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
         self.set_unit_status(self.build_unit_workload_status())
 
         if self._handle_potential_cluster_crash_scenario(state):
-            return
-
-        self._set_app_status(state)
+            self._set_app_status(state)
 
     def _set_app_status(self, state: str) -> None:
         """Set the application status based on the cluster state."""
--- a/kubernetes/tests/integration/helpers_ha.py
+++ b/kubernetes/tests/integration/helpers_ha.py
@@ -660,60 +660,6 @@ def stop_mysqld_service(juju: Juju, unit_name: str) -> None:
                 raise Exception("Failed to stop the mysqld process")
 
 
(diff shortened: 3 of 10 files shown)


kubernetes/src/charm.py
```diff
@@ -930,6 +930,30 @@ def _handle_potential_cluster_crash_scenario(self, state: str) -> bool:
...
+                self._mysql.stop_group_replication()
+                if self.unit.is_leader():
+                    # run on leader only for coordinate recovery
+                    logger.warning("Attempting reboot from complete outage")
+                    self._mysql.reboot_from_complete_outage()
+                    return False
+                else:
+                    # set offline state and delegate reboot from outage to leader
+                    logger.warning("Set instance to offline")
+                    self.unit_peer_data["member-state"] = InstanceState.OFFLINE
```
sinclert-canonical: Wait for resolution in this `8.0/edge` counterpart [comment](https://github.com/canonical/mysql-operators/pull/338#discussion_r3379119285).

machines/src/charm.py
```diff
@@ -438,11 +438,35 @@ def _charm_tracing_config(self) -> None:
...
+                self._mysql.stop_group_replication()
+                if self.unit.is_leader():
+                    # run on leader only for coordinate recovery
+                    logger.warning("Attempting reboot from complete outage")
+                    self._mysql.reboot_from_complete_outage()
+                    return False
+                else:
+                    # set offline state and delegate reboot from outage to leader
+                    logger.warning("Set instance to offline")
+                    self.unit_peer_data["member-state"] = InstanceState.OFFLINE
```
sinclert-canonical: Wait for resolution in this `8.0/edge` counterpart [comment](https://github.com/canonical/mysql-operators/pull/338#discussion_r3379121249).

astrojuanlu:
Thanks to Copilot I'm now realizing that on Kubernetes it's

https://github.com/canonical/mysql-operators/blob/a638747ed1c6eb03aa3f435e7120ada2b5eea4d1/kubernetes/src/charm.py#L1083-L1084

whereas on machines we have

https://github.com/canonical/mysql-operators/blob/a638747ed1c6eb03aa3f435e7120ada2b5eea4d1/machines/src/charm.py#L610-L611

(notice the **`not`**)

Any reason to have two almost identical methods with inverted logic?

kubernetes/src/charm.py
astrojuanlu: Deleting this part in the meantime, let's see how it works and what Paulo says.

kubernetes/src/charm.py
```diff
@@ -1081,9 +1104,7 @@ def _on_update_status(self, _: UpdateStatusEvent | None) -> None:
         self.set_unit_status(self.build_unit_workload_status())
 
         if self._handle_potential_cluster_crash_scenario(state):
-            return
-
-        self._set_app_status(state)
+            self._set_app_status(state)
```
astrojuanlu: @sinclert-canonical if you have a moment, would love to know your opinion on this (see the equivalent fragment further down for the machines charm). I think 1 of the 2 was wrong before.

kubernetes/src/relations/tls.py
```diff
@@ -125,23 +125,27 @@ def _get_peer_common_name(self) -> str:
         """Get a common name for the peer certificate attributes."""
         return self._get_common_name()
 
+    def _try_get_unit_address(self, relation_name: str) -> str:
+        """Get fqdn/address for a unit, or "" if DNS not propagated yet."""
+        try:
+            return self.charm.get_unit_address(self.charm.unit, relation_name)
+        except RuntimeError:
+            return ""
```
astrojuanlu: This method is to be used in the TLS module, whereas `get_unit_address` will retain its error handling capabilities for every other caller

Fixes #350

kubernetes/src/charm.py
```diff
@@ -360,14 +360,16 @@ def get_unit_hostname(self, unit_name: str | None = None) -> str:
...
+        Raises ``RuntimeError`` if the FQDN still cannot be resolved.
         """
+        unit_hostname = self.get_unit_hostname(unit.name)
         try:
-            unit_hostname = self.get_unit_hostname(unit.name)
             unit_dns_domain = get_k8s_fqdn(unit_hostname)
         except RuntimeError:
             logger.warning("Unit DNS domain name is not propagated yet")
-            return ""
+            raise
```
astrojuanlu: See above, fixes #350

kubernetes/tests/integration/integration/high_availability/test_primary_switchover.py
```diff
@@ -112,28 +112,29 @@ def test_cluster_failover_after_majority_loss(juju: Juju) -> None:
...
-            delay=15,
+    units_to_stop = [non_primary_units.pop(), primary_unit]
+
+    # ensure no update-status is triggered
+    with update_interval(juju, "30m"):
+        # SIGKILL mysqld on a majority of units
+        # so the survivor sees the killed units as UNREACHABLE
+        # rather than gracefully departed
+        for unit in units_to_stop:
+            force_kill_mysqld_service(juju, unit)
```
astrojuanlu: @paulomach I was unable to sustain quorum loss with this test when using Canonical K8s, so vibed this alternative approach that seems to work. If you agree (and the test also passes with MicroK8s), we can backport it to 8.0.

kubernetes/tests/integration/helpers_ha.py
```diff
@@ -660,6 +660,60 @@ def stop_mysqld_service(juju: Juju, unit_name: str) -> None:
                 raise Exception("Failed to stop the mysqld process")
 
 
+def force_kill_mysqld_service(juju: Juju, unit_name: str) -> None:
+    """SIGKILL mysqld and suppress Pebble's auto-restart.
```
astrojuanlu: You can see the unmistakable style of LLMs on this long docstring. I trimmed it down a bit and still considered the context is useful. But I'm okay compacting it a bit more.

kubernetes/tests/unit/conftest.py
```diff
@@ -24,6 +24,20 @@ def mock_refresh():
         yield
 
 
+@pytest.fixture(autouse=True)
+def mock_get_k8s_fqdn():
+    """Stub k8s DNS resolution so charm init does not hit the real resolver.
```
astrojuanlu, line 29: Only tangentially related, but should make unit tests more network-independent

machines/tests/integration/integration/high_availability/test_primary_switchover.py
```diff
@@ -111,27 +111,19 @@ def test_cluster_failover_after_majority_loss(juju: Juju) -> None:
...
-            ready=lambda status: all((
-                wait_for_unit_status(app_name, unit_to_promote, "active")(status),
-                wait_for_unit_status(app_name, units_to_kill[0], "maintenance")(status),
-                wait_for_unit_status(app_name, units_to_kill[1], "maintenance")(status),
-            )),
-            timeout=15 * MINUTE_SECS,
-            delay=15,
+    # ensure no update-status is triggered
+    with update_interval(juju, "30m"):
+        subprocess.run(["lxc", "restart", "--force", machine_name[0], machine_name[1]], check=True)
```
astrojuanlu: I haven't backported the approach to the machines test yet, happy to do so if desired.

kubernetes/tests/integration/integration/high_availability/test_primary_switchover.py
paulomach: I like it, but there's more.

The use case for force_quorum is when other units are unreachable and stays like that, hence the `with update_interval(juju, "30m")` that prevents self healing.

When don't even need to care about them rejoining the cluster, but just assert that the surviving one is a primary.

On that sense, we can get rid of the lines 133-137 also.

kubernetes/tests/integration/integration/high_availability/test_primary_switchover.py
```diff
@@ -144,19 +145,3 @@ def test_cluster_failover_after_majority_loss(juju: Juju) -> None:
         )
```
paulomach: We don't need those

machines/tests/integration/integration/high_availability/test_primary_switchover.py
paulomach: We can stop the machines and f*** off, similarly to k8s approach

kubernetes/tests/integration/integration/high_availability/test_primary_switchover.py
paulomach: Check [this commit on 8.0 PR](https://github.com/canonical/mysql-operators/pull/338/changes/b4c1a030c18afe0d4f1b460ed29555b4adaaa850)

astrojuanlu: Excellent. Ported those changes here as well.

kubernetes/tests/integration/helpers_ha.py
astrojuanlu: Removed excessive docs.

machines/tests/integration/integration/high_availability/test_primary_switchover.py
astrojuanlu: Ported here!

kubernetes/src/relations/tls.py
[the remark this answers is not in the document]
astrojuanlu: Ughhhhh ugly as hell. But the insight is probably correct. Will look for an alternative fix.

astrojuanlu:
This PR is starting to drift from #338. So I'll spin off the fixes for empty connection details to a separate PR to reduce scope.

machines/tests/integration/integration/high_availability/test_self_healing_safe_recover_quorum.py
```diff
@@ -0,0 +1,97 @@
...
+
+    non_primary_units = app_units - {primary_unit}
+
+    unit_to_survive = non_primary_units.pop()
+
+    logging.info("Simulate quorum loss")
+    logging.info(f"Unit selected for survival: {unit_to_survive}")
+
+    units_to_kill = [non_primary_units.pop(), primary_unit]
+    kill_units(juju, app_name, units_to_kill)
```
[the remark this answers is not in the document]
astrojuanlu: This is killing the primary and a secondary, always. So I don't think this is flaky (or at least, it shouldn't).

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -2312,6 +2317,13 @@ def start_group_replication(self) -> None:
         with suppress(ExecutionError):
             self._instance_client_tcp.start_instance_replication()
 
+    def is_cluster_in_no_quorum(self) -> bool:
+        """Check whether the cluster has lost quorum as seen from this instance."""
+        status = self.get_cluster_status()
+        if not status:
+            return False
+        return status["defaultReplicaSet"]["status"] == ClusterStatus.NO_QUORUM
```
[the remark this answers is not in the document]
astrojuanlu: `defaultReplicaSet` is always assigned

https://github.com/mysql/mysql-shell/blob/71f4e4a0ddb86d37bae838321720983e5035764e/modules/adminapi/cluster/status.cc#L2540-L2541

And so is `status`

https://github.com/mysql/mysql-shell/blob/71f4e4a0ddb86d37bae838321720983e5035764e/modules/adminapi/cluster/status.cc#L1839-L1881

So, no.

kubernetes/tests/unit/test_charm.py
```diff
@@ -312,6 +312,24 @@ def test_get_unit_address(self, mock_get_k8s_fqdn):
...
         mock_get_k8s_fqdn.assert_called_once_with("mysql-k8s-0.mysql-k8s-endpoints")
 
+    @patch(
+        "charm.get_k8s_fqdn",
+        side_effect=[
+            RuntimeError("Failed to resolve canonical name for mysql-k8s-0.mysql-k8s-endpoints"),
+            "mysql-k8s-0.mysql-k8s-endpoints.default.svc.cluster.local",
+        ],
+    )
+    def test_get_unit_address_retries_on_dns_not_propagated(self, mock_get_k8s_fqdn):
```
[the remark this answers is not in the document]
astrojuanlu: The mock didn't work and I will not bother for 2 seconds.

kubernetes/src/charm.py
```diff
@@ -1084,9 +1109,7 @@ def _on_update_status(self, _: UpdateStatusEvent | None) -> None:
         self.set_unit_status(self.build_unit_workload_status())
 
         if self._handle_potential_cluster_crash_scenario(state):
-            return
-
-        self._set_app_status(state)
+            self._set_app_status(state)
```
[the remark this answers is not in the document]
astrojuanlu: I will try to avoid having to level the logic of both charms. For now, I hope it is enough to avoid introducing regressions.

kubernetes/src/relations/tls.py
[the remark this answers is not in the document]
astrojuanlu: Fixed in https://github.com/canonical/mysql-operators/pull/364

kubernetes/src/charm.py
astrojuanlu: On second thought, I think we should review this in another PR. For now, I just changed from

```
if <bool>:
    return

# else
do_sth()
```

to

```
if not <bool>:
    do_sth()
```

which is more compact and clear.

machines/src/charm.py
```diff
@@ -442,11 +442,37 @@ def _charm_tracing_config(self) -> None:
...
+                self._mysql.stop_group_replication()
+                if self.unit.is_leader():
+                    # run on leader only for coordinate recovery
+                    logger.warning("Attempting reboot from complete outage")
+                    self._mysql.reboot_from_complete_outage()
+                    return False
+            except MySQLRebootFromCompleteOutageError:
+                logger.error("Failed to reboot cluster from complete outage")
+                self.set_unit_status(BlockedStatus("failed to recover cluster"))
+            return True
```
astrojuanlu, line 477: I find it suspicious that the handler logic is inverted between machines and k8s, and yet the `return`s in this block are the same.

astrojuanlu:
Despite this PR intending to make fixes to no-quorum situations, both there (8.0/edge) and here (8.4/edge) `test_primary_switchover.py` is _sometimes_ failing. In fact, it failed in one of the architectures in the last commit of #338 before getting merged.

The problem seems to be that the self-healing service introduced in #342 and #343 _sometimes_ runs right in the middle of the quorum loss, stops group replication, unit becomes offline, and promote-to-primary fails.

Logs:
- 8.0/edge (#338 right before merging) https://github.com/canonical/mysql-operators/actions/runs/27357958065/job/80847830020
- 8.4/edge (this PR, last commit) https://github.com/canonical/mysql-operators/actions/runs/27411952089/job/81019475171?pr=354

In both cases:

```
This function is not available through a session to an InnoDB Cluster that belongs to an InnoDB ClusterSet but is not ONLINE
```

Ironically, in the cases in which the test passed, we technically did create a no quorum situation (we stopped mysqld in a majority of nodes, including the primary) but the secondary never noticed because the status update never triggered (on purpose).

We didn't include any provision to slow down or disable the self-healing system, so disabling it for our tests requires some hacks.

---

The rest of the failures are flakiness.

paulomach:
> Despite this PR intending to make fixes to no-quorum situations, both there (8.0/edge) and here (8.4/edge) `test_primary_switchover.py` is _sometimes_ failing. In fact, it failed in one of the architectures in the last commit of #338 before getting merged.
> 
> The problem seems to be that the self-healing service introduced in #342 and #343 _sometimes_ runs right in the middle of the quorum loss, stops group replication, unit becomes offline, and promote-to-primary fails.
> 

I think you nailed the diagnosis @astrojuanlu 
For clarity, with the auto `no_quorum` safe recovery, we are making an implicit assumption that might be false: all nodes are reachable. If they are, the cluster can recover with the `reboot_from_complete_outage` path.

But when not all nodes are reachable, that's when we need the `promote-to-primary force=true` path. That will recover regardless of all nodes being available (e.g. cluster of 3, where 2 nodes crashed, but only 1 recovers).

I'll propose a PR to make the assumption (all nodes are reachable) explicit.

astrojuanlu:
All `test_primary_switchover.py` passing 👍🏼

paulomach (approved):

kubernetes/src/charm.py
```diff
@@ -1083,10 +1125,10 @@ def _on_update_status(self, _: UpdateStatusEvent | None) -> None:
         self.unit_peer_data["member-state"] = state
         self.set_unit_status(self.build_unit_workload_status())
 
-        if self._handle_potential_cluster_crash_scenario(state):
-            return
-
-        self._set_app_status(state)
+        # TODO: Logic here is almost the opposite as the machines charm, but not quite
+        # We should review and fix it
```
sinclert-canonical, line 1144: Was the investigation completed? Anything we could simplify before merging?

kubernetes/tests/integration/helpers_ha.py
```diff
@@ -699,3 +699,37 @@ def create_app_secret(
...
+
+
+def force_kill_mysqld_service(juju: Juju, unit_name: str) -> None:
+    """SIGKILL mysqld and suppress Pebble's auto-restart.
+
+    Args:
+        juju: The Juju model.
+        unit_name: The name of the unit.
+
+    """
```
sinclert-canonical, line 711: Similar to the `8.0/edge` counter-part [PR](https://github.com/canonical/mysql-operators/pull/338):

These operations seems similar to what we do for the process-kill integration test (see [source](https://github.com/canonical/mysql-operators/blob/8720ea4b88fc4886bc9d33fa4292f672af4f306e/kubernetes/tests/integration/integration/high_availability/test_self_healing_process_killed.py#L84-L91)). In the future, we can centralize them to a single method + analyze whether we can ditch `lightkube` in all the integration test use-cases. Let's leave it in-sync with `8.0/edge` for now.

sinclert-canonical (approved):

kubernetes/src/charm.py
astrojuanlu: Not really. I would have flipped the logic if it was 100 % reversed, but this new block is not. I'd like to defer that to a future PR.

kubernetes/tests/integration/helpers_ha.py
astrojuanlu: Happy to leave a TODO as well so we don't forget

astrojuanlu: Done.

sinclert-canonical:
Just noticed: this PR never included the spread tasks for the integration tests to run.

astrojuanlu:
Neither did #338 😓

PR #345, opened 2026-06-04
[DPE-10337] Use non-stale data node for cluster belonging

paulomach:
## Issue

See #344

## Solution

- Use metadata from active/primary unit to check instance in the cluster
- Bonus: k8s charm was not setting unknown|recovering unit status

Fix #344

PR #356, opened 2026-06-09
[DPE-10337] Use non-stale data node for cluster belonging(8.4)

paulomach:
Port PR #345

### DIFF
--- a/kubernetes/src/charm.py
+++ b/kubernetes/src/charm.py
@@ -1040,7 +1040,7 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
         if member_state == "UNKNOWN" or member_state == InstanceState.RECOVERING:
             # avoid changing status while tls is being set up or charm is being initialized
             logger.info(f"Unit {member_state=}")
-            self.unit.status = MaintenanceStatus(member_state)
+            self.set_unit_status(MaintenanceStatus(member_state))
             return True
 
         # avoid changing status while async replication is setting up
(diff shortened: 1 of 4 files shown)


kubernetes/src/charm.py
```diff
@@ -1037,6 +1040,7 @@ def _is_cluster_blocked(self) -> bool:
         if member_state == "UNKNOWN" or member_state == InstanceState.RECOVERING:
             # avoid changing status while tls is being set up or charm is being initialized
             logger.info(f"Unit {member_state=}")
+            self.unit.status = MaintenanceStatus(member_state)
```
astrojuanlu: It's unclear to me whether we should always use `self.set_unit_status` 🤔

astrojuanlu (approved):
Approved, but left a comment

kubernetes/src/charm.py
dragomirp: I think it's necessary for refreshv3, since it runs on every hook.

paulomach: Good catch

3197a3574

sinclert-canonical: Drago is right here. See refresh V3 [status docs](https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/status/).

Refresh V3 (Charmed MySQL 8.4 only) sets some app / unit statuses which should take higher priority compared to any the ones set manually by the operator code. IIRC, the charm initialization part is the only one where `set_unit_status` should not be called.

sinclert-canonical (approved):

PR #374, opened 2026-06-16
[misc] (8.0/edge) fixes for relations tests

paulomach:
fixes for k8s relation tests:
- test_mysql_root.py
- test_database.py
___
- Usage of secrets using ops and failing
- Some waits criteria were not working

PR #378, opened 2026-06-18
[MISC] Revert use of Juju binary to detect version version

astrojuanlu:
## Issue

This helps publish the default Allure results when Juju is not installed, avoiding https://github.com/canonical/mysql-operators/actions/runs/27763437351/job/82152214162#step:5:147

## Checklist

### DIFF
--- a/kubernetes/tests/integration/juju_.py
+++ b/kubernetes/tests/integration/juju_.py
@@ -1,9 +1,13 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-import jubilant_backports
+import importlib.metadata
 
-juju = jubilant_backports.Juju()
+import ops
 
-has_secrets = not juju._is_juju_2
-juju_major_version = 2 if juju._is_juju_2 else 3
+# libjuju version != juju agent version, but the major version should be identical—which is good
+# enough to check for secrets
+_libjuju_version = importlib.metadata.version("juju")
+has_secrets = ops.JujuVersion(_libjuju_version).has_secrets
+
+juju_major_version = int(_libjuju_version.split(".")[0])
--- a/machines/tests/integration/juju_.py
+++ b/machines/tests/integration/juju_.py
@@ -1,9 +1,13 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-import jubilant_backports
+import importlib.metadata
 
-juju = jubilant_backports.Juju()
+import ops
 
-has_secrets = not juju._is_juju_2
-juju_major_version = 2 if juju._is_juju_2 else 3
+# libjuju version != juju agent version, but the major version should be identical—which is good
+# enough to check for secrets
+_libjuju_version = importlib.metadata.version("juju")
+has_secrets = ops.JujuVersion(_libjuju_version).has_secrets
+
+juju_major_version = int(_libjuju_version.split(".")[0])


paulomach (approved):

sinclert-canonical (commented):
Why don't we revert the changes introduced in PR https://github.com/canonical/mysql-operators/pull/374, and leave the `juju_.py` modules as they were?

```python
import importlib.metadata

import ops

# libjuju version != juju agent version, but the major version should be identical—which is good
# enough to check for secrets
libjuju_version = importlib.metadata.version("juju")

has_secrets = ops.JujuVersion(_libjuju_version).has_secrets
juju_major_version = int(_libjuju_version.split(".")[0])
```

paulomach:
> Why don't we revert the changes introduced in PR #374, and leave the `juju_.py` modules as they were?
> 
> ```python
> import importlib.metadata
> 
> import ops
> 
> # libjuju version != juju agent version, but the major version should be identical—which is good
> # enough to check for secrets
> libjuju_version = importlib.metadata.version("juju")
> 
> has_secrets = ops.JujuVersion(_libjuju_version).has_secrets
> juju_major_version = int(_libjuju_version.split(".")[0])
> ```



> Why don't we revert the changes introduced in PR #374, and leave the `juju_.py` modules as they were?
> 

Because that was giving us the wrong response, ~~as we stop installing libjuju-2.x since some change in the workflow. I'll track it down.~~ that was hidden in [spread](https://github.com/canonical/mysql-operators/blob/1e394788d3ebc78ab1168ce685aa0f64a363b793/machines/spread.yaml#L141) 
I don't know why it was not working now. The point is, tests relying in `has_secrets` where failing as it was always `True`

sinclert-canonical:
I do not understand why we are not just reverting the changes.

---

I see that nightly runs for the `8.0/edge` branch are crashing. Sure, that should not happen. However, that started happening because of the changes introduced in PR https://github.com/canonical/mysql-operators/pull/374, because unless a Juju environment is bootstraped in the collect-integration-tests [CI job](https://github.com/canonical/mysql-operators/blob/1e394788d3ebc78ab1168ce685aa0f64a363b793/.github/workflows/integration_test.yaml#L21-L92), `jubilant-backports` is not able to determine the version of Juju installed.

However, I fail to understand what was wrong before. I thought the referenced [snippet](https://github.com/canonical/mysql-operators/blob/1e394788d3ebc78ab1168ce685aa0f64a363b793/machines/spread.yaml#L141) is only applicable to the integration tests themselves, and not to the collect-integration-test job. What was broken before? Do we have specific nightly run examples where we see version `2.X.Y` of the Juju python library getting installed in Juju 3 integration tests?

PostgreSQL VM 14 and PostgreSQL K8s 14 seems to have the exact same setup that we used to have, and AFAIK they are not seeing those errors.

astrojuanlu:
I submitted the PR as a partial revert, but maybe a better solution would be to do a `if Juju 2: poetry ...` in the collect-jobs test, and even remove that logic from spread. Spread is better suited for detecting the "variant" but relying on that code is what brought us to this issue, I think.

paulomach:
> I submitted the PR as a partial revert, but maybe a better solution would be to do a `if Juju 2: poetry ...` in the collect-jobs test, and even remove that logic from spread. Spread is better suited for detecting the "variant" but relying on that code is what brought us to this issue, I think.

Definitely revert it for now/
I missed that completely while debugging the test, because we do not tend to use spread on manual testing.

astrojuanlu:
Reverted 👍🏼

sinclert-canonical (approved):
Apologies for the back-and-forth, but reverting is the sanest approach here.

Please share an example of what was broken before, whenever you have the chance. That way I can at least understand what we were aiming to solved with the initial PR proposal.

PR #379, opened 2026-06-18
[MISC] (8.4/edge) Fix full cluster recovery when async related

paulomach:
Port PR #332 to 8.4/edge


sinclert-canonical (approved):
This should be the last 8.0 <-> 8.4 bug fix left to be ported, according to my book.

astrojuanlu (approved):

paulomach:
> This should be the last 8.0 <-> 8.4 bug fix left to be ported, according to my book.

Your book is my book - so yes. We can run an llm through git history to be sure.

PR #382, opened 2026-06-19
[MISC] Ease up self-healing startup checks

paulomach:
## Issue

We are not starting the `self_healing` manager early on after crash (pod/node or host).
The in place checks prevent the background process to be started (e.g. waiting for `ActiveStatus`), defeating the purpose of having a faster, `update_status` independent custom hook for such cases.

## Solution

Given that most guardrails for self-healing are already coded in the `_on_update_status` hook handler, we can allow the background process that periodically call it to start early on host/pod recovery.

- **prevent uncaught KeyError** for early calls of `unit_initialized` 
- **relaxed self_healing background process start criteria**

_Extra_:
- consider removing **all** update_status interval on tests 
- consider speed up self_healing interval (currently at 120s)

### DIFF
--- a/kubernetes/lib/charms/mysql/v0/mysql.py
+++ b/kubernetes/lib/charms/mysql/v0/mysql.py
@@ -851,6 +851,9 @@ class MySQLCharmBase(CharmBase, ABC):
 
     def unit_initialized(self, raise_exceptions: bool = False) -> bool:
         """Check if the unit is added to the cluster."""
+        if not self.app_peer_data.get("cluster-name"):
+            # early calls - cluster name not set, avoid KeyError
+            return False
         try:
             return self._mysql.cluster_metadata_exists()
         except MySQLClusterMetadataExistsError:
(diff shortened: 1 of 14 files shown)


kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -851,6 +851,9 @@ def removing_unit(self) -> bool:
 
     def unit_initialized(self, raise_exceptions: bool = False) -> bool:
         """Check if the unit is added to the cluster."""
+        if not self.app_peer_data.get("cluster-name"):
+            # early calls - cluster name not set, avoid KeyError
+            return False
```
sinclert-canonical, line 856: I wish there was a better check for whether the unit is fully initialized, but alas, this is what we got.

machines/lib/charms/mysql/v0/mysql.py
```diff
@@ -850,6 +850,9 @@ def removing_unit(self) -> bool:
 
     def unit_initialized(self, raise_exceptions: bool = False) -> bool:
         """Check if the unit is added to the cluster."""
+        if not self.app_peer_data.get("cluster-name"):
+            # early calls - cluster name not set, avoid KeyError
+            return False
```
sinclert-canonical, line 855: I wish there was a better check for whether the unit is fully initialized, but alas, this is what we got.

sinclert-canonical (approved):

astrojuanlu (approved):

astrojuanlu:
> _Extra_:
> 
> * consider removing **all** update_status interval on tests
> * consider speed up self_healing interval (currently at 120s)

+1!

PR #388, opened 2026-06-27
[MISC] preserve type on config dict

paulomach:
## Issue

ConfigParser will stringfy all values, when converting back to dict, original types are lost.
Some values need to carry their type so they are correctly quoted on mysql-shell-client.
E.g.: `experimental-max-connections` is a int, but will be quoted as string, and will break when applying in the instance: `set global max_connections='200' != set global max_connections=200`

## Solution

Instead of rendering the configuration as a ConfigParser, render in dict and convert to CP when writing to file.

### DIFF
--- a/kubernetes/lib/charms/mysql/v0/mysql.py
+++ b/kubernetes/lib/charms/mysql/v0/mysql.py
@@ -1161,12 +1161,7 @@ class MySQLBase(ABC):
                 performance_schema_instrument = "'memory/%=OFF'"
 
         binlog_retention_seconds = binlog_retention_days * 24 * 60 * 60
-        config = configparser.ConfigParser(interpolation=None)
-
-        # do not enable slow query logs, but specify a log file path in case
-        # the admin enables them manually
-        config["mysqld"] = {
-            # All interfaces bind expected
+        mysqld_config: dict = {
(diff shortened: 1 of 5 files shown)


kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -1179,36 +1174,39 @@ def render_mysqld_configuration(  # noqa: C901
             "general_log_file": f"{snap_common}/var/log/mysql/general.log",
             "loose-group_replication_paxos_single_leader": "ON",
             "slow_query_log_file": f"{snap_common}/var/log/mysql/slow.log",
-            "binlog_expire_logs_seconds": f"{binlog_retention_seconds}",
+            "binlog_expire_logs_seconds": binlog_retention_seconds,
```
sinclert-canonical, line 1177: Why to transform this integer into a string? I thought we were doing the other way around.

paulomach: I don't get it. It's avoiding conversion.

sinclert-canonical: My bad, I understood it the other way around 🤪

PR #391, opened 2026-06-30
[DPE-10023] feat: enable metrics exporter by default

Soundarya03:
Introduces changes to machine and k8s operators to enable metrics exporter by default, rather than only upon COS integration.

### DIFF
--- a/kubernetes/lib/charms/mysql/v0/mysql.py
+++ b/kubernetes/lib/charms/mysql/v0/mysql.py
@@ -573,7 +573,7 @@ class MySQLCharmBase(CharmBase, ABC):
 
         self.set_secret("app", secret_key, new_password)
 
-        if username == MONITORING_USERNAME and self.has_cos_relation:
+        if username == MONITORING_USERNAME:
             self._mysql.restart_mysql_exporter()
 
     def _get_cluster_status(self, event: ActionEvent) -> None:
@@ -3102,7 +3102,7 @@ class MySQLBase(ABC):
     @abstractmethod
     def restart_mysql_exporter(self) -> None:
         """Restart the mysqld exporter."""
-        raise NotImplementedError
+        raise NotImplementedError # bookmark
 
     @abstractmethod
     def wait_until_mysql_connection(self, check_port: bool = True) -> None:
--- a/machines/tests/unit/test_charm_base.py
+++ b/machines/tests/unit/test_charm_base.py
@@ -3,7 +3,7 @@
 
 import unittest
 from pathlib import Path
-from unittest.mock import patch
+from unittest.mock import Mock, PropertyMock, patch
 
 from charms.mysql.v0.mysql import MySQLCharmBase, MySQLSecretError
 from ops.model import RelationDataTypeError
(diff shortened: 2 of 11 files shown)


kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -3102,7 +3102,7 @@ def start_mysqld(self) -> None:
     @abstractmethod
     def restart_mysql_exporter(self) -> None:
         """Restart the mysqld exporter."""
-        raise NotImplementedError
+        raise NotImplementedError # bookmark
```
paulomach: left over?

kubernetes/lib/charms/mysql/v0/mysql.py
```diff
@@ -573,7 +573,7 @@ def _on_set_password(self, event: ActionEvent) -> None:
 
         self.set_secret("app", secret_key, new_password)
 
-        if username == MONITORING_USERNAME and self.has_cos_relation:
```
paulomach, line 576: Please check if we can get rid of the `self.has_cos_relation` altogether also.

paulomach (commented):
@Soundarya03 , couple comments for k8s charm which are valid for machines. Looks good otherwise

kubernetes/lib/charms/mysql/v0/mysql.py
Soundarya03: Yes, sorry. Cleaned up now!

kubernetes/lib/charms/mysql/v0/mysql.py
Soundarya03: It seems to be used in log rotation compression, although yes, this means it is used only in one place now. 

I still think the we should keep the method though. What do you think?

machines/tests/unit/test_charm_base.py
```diff
@@ -128,3 +128,43 @@ def test_abstract_methods(self):
...
+        )
+
+        self.harness.run_action(
+            "set-password", {"username": "charmed-stats", "password": "newpass123"}
+        )
+
+        _mysql.return_value.update_user_password.assert_called_once_with(
+            "charmed-stats", "newpass123"
+        )
+        _mysql.return_value.restart_mysql_exporter.assert_called_once()
```
sinclert-canonical: Why `_mysql.return_value` and not just `_mysql` like other tests within this module?

machines/tests/unit/test_charm_base.py
```diff
@@ -128,3 +128,43 @@ def test_abstract_methods(self):
...
+        )
+
+        self.harness.run_action(
+            "set-password", {"username": "charmed-operator", "password": "newpass123"}
+        )
+
+        _mysql.return_value.update_user_password.assert_called_once_with(
+            "charmed-operator", "newpass123"
+        )
+        _mysql.return_value.restart_mysql_exporter.assert_not_called()
```
sinclert-canonical: Why `_mysql.return_value` and not just `_mysql` like other tests within this module?

machines/tests/unit/test_charm_base.py
Soundarya03: A slight difference in the patching, corrected it to follow the pattern followed by others now!

kubernetes/lib/charms/mysql/v0/mysql.py
paulomach: sure, thanks for checking

paulomach:
> lgtm: before merge let's talk about the note made about the increase in attack surface area.

Per threat model, we don't asses the exporter being available conditionally - so it's modeled as always available with acceptable risk level.

So let's go ahead with merging.

